PR #3, opened 2026-07-23
feat: match os stopwatch

cypressru:
Now a three-commit stack: sram → sync → stopwatch (same content as the pr-ossram and pr-ossync PRs — the diff shrinks as those merge). All five stopwatch functions at 100% under their SDK names, shipped Matching, and the rebuilt main.dol SHA-1 matches retail (`9214426b…04a8`) with all three objects linked in.

Why the stack: the earlier draft hung mwldeppc during layout. Root cause (bisected, reproduces under both wibo and wine, so it's the real 2003 linker): linking a compiled OSStopwatch.o while the syscall-vector region after it is an auto-split object wedges the linker, and the sram unit's old range — which sliced through this TU — contributed. With corrected sram boundaries (0x801D541C–0x801D60C0) and the sync unit configured, it links in seconds. The stopwatch functions carry SDK names with global scope because the engine's profiling code calls them (fn_800CDE58, fn_80112BB0).

Also adds the OSGetTime prototype to os.h, since a second TU now calls it.

Verified with a fresh full build: objdiff 100% on every function, object section sizes byte-compared against the extracted originals.


cypressru:
Converting to draft. Found a linker interaction while verifying this branch against current main: with this unit added, mwldeppc hangs during layout on a full local build. The functions themselves verify at 100% in objdiff; bisecting the interaction now, will mark ready when the branch links clean. Please hold off pulling this locally until then.

Jovinull:
Awesome work getting them to 100%! Thanks for the heads up on the linker hang. Take your time bisecting it, I'll leave this on standby until you give the green light that the build is clean.

cypressru:
Resolved — the linker hang is root-caused and fixed (details in the updated description). Safe to pull and build locally now; a fresh full build verifies the DOL SHA-1 end-to-end.

PR #10, opened 2026-07-23
feat: match os sync

cypressru:
Adds OSSync.c (system call vector installation): both functions at 100%, .text 132/132, Matching, DOL verifies. The vector body is assembly in the original source too — it runs as the sc handler and returns with rfi, which C can't express. Configuring this unit also removes a degenerate auto-split object at 0x801D628C that the stopwatch PR #3 depends on.


Jovinull:
Closing as superseded — no work is being lost here.

The OSSync work this PR carries already landed on `main` in 175ea03 (via #3). I verified this by diffing the two directly: `src/dolphin/os/OSSync.c` is byte-for-byte identical between this branch and `main`, and every other line the PR adds is already present:

- `SystemCallVector = .text:0x801D628C` in `symbols.txt`
- the `dolphin/os/OSSync.c` block in `splits.txt`
- `Object(Matching, "dolphin/os/OSSync.c")` in `configure.py`

The conflict shown was purely an artifact of the branch predating the merge order fix, not a real divergence. `OSSync` is reported complete on `main` and the DOL still hashes to 9214426b8a3fb1d6fe3dcff09bcc1a959e1e04a8.

Thanks for the work — it is in, just under the other commit.

PR #26, opened 2026-07-23
refactor: reach the exi registers by pointer bump in dbcomm

Jovinull:
**This does not complete any function. Nothing in `dbcomm.c` reaches 100%, so the object stays `NonMatching`.** It is a large step toward the match and a correction to what the file's header comment blamed.

## What the original actually does

The header used to attribute five of the unmatched functions to a vague allocator decision. The real cause is addressing: the original reaches the channel 2 registers by *bumping a pointer*, not by indexing. `stwu` and `lwzu` only come out of `*(p += n)`; `EXI_REGS[n]` folds the offset into a displacement and rebuilds the address at every use.

Two details go with it:

* DBWrite derives `csr` and `cr` inside its **first** poll block only; the three later blocks reuse what that block left bumped. Bumping in every block instead costs 1.7%.
* Declaration order decides which callee saved register a local lands in. Declaring `word` before `i` in `fn_801F8988`, and `err` before the buffer pointer in the block helpers, puts them where the original has them.

## Result

| function | before | after |
| --- | --- | --- |
| `fn_801F8724` | 64.09% | 94.65% |
| `fn_801F8800` | 64.09% | 94.65% |
| `fn_801F8678` | 72.70% | 90.56% |
| `fn_801F88DC` | 72.70% | 90.56% |
| `DBWrite` | 86.53% | 92.51% |
| `fn_801F8988` | 87.55% | 88.55% |

`DBInitComm` (93.3%) and `EXIHandler` (87.5%) are untouched. SDK fuzzy 89.33% -> 89.73%.

`main.dol` still hashes to `9214426b8a3fb1d6fe3dcff09bcc1a959e1e04a8`, verified on a from-scratch rebuild with `build/` wiped.

## What is left

One thing, and it is the same thing in every function: the original reserves more stack below its locals than we do. Frames run 0x08 over on the two block helpers, 0x10 on the two read helpers and 0x30 on DBWrite, and every remaining difference is that offset showing up in the prologue, the epilogue and each stack reference. Registers, instruction order and instruction count already agree. Part of DBWrite's share is visible: the original keeps five distinct slots for the command and trailer words at 0x3c, 0x44, 0x4c, 0x50 and 0x54, where ours collapses the three command words onto one because their live ranges do not overlap.

Measured and rejected, recorded in the header so they are not retried: `-use_lmw_stmw on`, `-inline off`, `-inline noauto`, `-O3,p`, `-O4`, `-O4,s` (breaks a matching function), `-opt nopeephole`, `-opt nostrength`, `-opt noschedule`, `-opt nolifetimes`, `-common on`, `-str reuse,pool,readonly`, `-sdata 0 -sdata2 0`; dropping `static` off the helpers; splitting `i = 0` out of the for init (-0.06%); giving `fn_801F8988` its own control register pointer local (-1.2%); and reading `DBCommHandler` into a local in `EXIHandler`.

## AI assistance

Written with Claude Code. Every number above comes from objdiff, and the DOL hash was checked on a clean rebuild.

### DIFF
--- a/src/dolphin/db/dbcomm.c
+++ b/src/dolphin/db/dbcomm.c
@@ -20,46 +20,51 @@
 // right order and the right count; what separates them from the original is
 // where values live, not what the code does.
 //
-// Four of the five share one cause. The original keeps the addresses of the EXI
-// registers in callee saved registers for the whole body, which is why it saves
-// six of them with stmw and forms the addresses with stwu and lwzu. Ours holds
-// fewer and rebuilds each address from the constant, because a constant address
-// is cheap to rematerialize and the compiler would rather do that than spend a
-// register. Nothing tried so far changes that decision: absolute macros, an
-// indexed macro off a base, a file scope const pointer, locals holding the two
-// registers, and a local holding the base were all measured. Two findings did
-// come out of it. The register block has to be volatile or the reads get folded
-// away, worth nine points on the read helpers. And holding the registers in
-// locals helps some functions and hurts others, worth three points on the read
-// helpers and minus two on fn_801F8988, so it is not a general answer.
+// The cause that used to be shared by five of them is solved. The original
+// reaches the two channel 2 registers by bumping a pointer, not by indexing:
+// stwu and lwzu only come out of *(p += n), while EXI_REGS[n] folds the offset
+// into a displacement and rebuilds the address at every use. Writing the bump
+// took fn_801F8724 and fn_801F8800 from 64.1% to 93.9%, both read helpers from
+// 72.7% to 90.6% and DBWrite from 86.5% to 92.5%. Two details go with it. The
+// register block still has to be volatile or the reads fold away. And the bump
+// belongs only where the original put it: DBWrite derives csr and cr inside its
+// first poll block and the three later blocks reuse what that block left
+// bumped, so bumping in every block costs 1.7%, and giving fn_801F8988 its own
+// pointer local for the control register costs 1.2% because the extra live
+// value pushes the two payload loops off their registers.
 //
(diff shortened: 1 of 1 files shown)


Jovinull:
Kept digging on the frame after opening this. Four more hypotheses eliminated, all recorded in the file header so they are not retried:

* **Compiler version.** 1.2.5n is right and this file is not part of the debugger library despite what it talks to. Averaged over the twelve functions: 1.2.5n 94.4% with four exact, 1.2.5 91.6% with two, 1.3 and 1.3.2 both collapse to ~52% with none.
* **Parameter save area.** Giving `fn_801F8988` three more parameters leaves every caller's frame byte for byte where it was, so this compiler does not size that area by argument count while the arguments still fit in registers.
* **Compiler options.** Thirteen swept, none move the frame; `-O4,s` actively breaks a function that already matches.
* **Linkage.** Dropping `static` off the helpers changes nothing, though symbols.txt carries no `scope:local` marker for any of them.

That narrows it to one thing. Every `r1` relative access in the original is accounted for — `fn_801F8678` touches one word at 0x18, `fn_801F8724` two at 0x20 and 0x24 — and each still reserves twenty four bytes it never reads or writes. `DBRead` settles what that space is not: it matches exactly, it makes a three argument call like the block helpers do, and it reserves twelve bytes, so the parameter save area we emit is already the right size.

The surplus is locals the original declared and never used. They leave no trace in the instruction stream, so short of the original source or a reference decompilation of this file there is nothing left to infer them from. Registers, instruction order and instruction count already agree in every function here.

Still non-matching, unchanged from the description above.

PR #34, opened 2026-07-23
feat: match os context

cypressru:
Adds OS context creation, save/load and switching, floating-point context management, context diagnostics, and exception-handler initialization.

The register and floating-point save/restore paths remain PowerPC assembly because they directly manipulate the complete register file, special-purpose registers, paired-single state, MSR, and exception return state. The ordinary context-management and diagnostic routines remain in C. One matching-sensitive detail is that the saved FPU state mask must be unsigned; this preserves the original unsigned comparison in OSDumpContext.


Jovinull:
Merged. Conflict against OSError.c, which landed first: kept both objects in `splits.txt`/`configure.py`, and since this PR renames `fn_801D2770` to `OSGetStackPointer`, I declared it in `include/dolphin/os.h` and updated the call site in `OSPanic`. Rebuilt clean: SHA-1 gate 18/18, OSContext.c 13/13 at 100%, OSError.c still 100%.

PR #37, opened 2026-07-23
feat: match dvd id utils

cypressru:
Adds the DVD disk-ID comparison helper, including wildcard handling for game names, disk numbers, and game versions.

The function was verified byte-for-byte in objdiff, and the object passed the fresh-build, section-size, Matching-link, DOL SHA-1, and all 17 REL SHA-1 gates.

One boundary-sensitive detail is that the following 0x30-byte function is not DVDGenerateDiskID, despite occupying that address in later SDK revisions. It belongs to the separate DVD fatal-message translation unit, so this object ends at 0x801DBA9C.


Jovinull:
Merged. Conflict against #36 (dvdFatal), which landed first — both PRs name the same range from opposite ends and agree on the split at 0x801DBA9C, so the resolution keeps `DVDCompareDiskID` from this PR and `__DVDPrintFatalMessage` from #36. Rebuilt clean: SHA-1 gate 18/18, both objects 100%.

PR #38, opened 2026-07-23
feat: match fst load

cypressru:
Adds the boot-time FST loader, including disk-ID reads, boot-block loading, FST transfer, arena adjustment, and disc metadata reporting.

One matching-sensitive detail is that the asynchronous boot-sector callback uses C long, rather than the project’s typedef-based s32; both are 32-bit on this target, but the distinct type controls callback compatibility. The TU also owns the aligned small-data slice containing its newline and ON/OFF strings; assigning that slice correctly is required to preserve the final DOL and REL layouts.


Jovinull:
Merged. Conflict against #36 (dvdFatal) and #37 (dvdidutils), which landed first. The substantive one was the `.sbss` block at 0x8042CC60 — this PR still saw it as a single 8-byte symbol, while #36 had already split `FatalFunc` out of the first four bytes. Resolution keeps #36's split and takes your `scope:local` on `lbl_8042CC68`, which fstload owns. Rebuilt clean: SHA-1 gate 18/18, fstload.c 2/2 at 100%, neighbours still 100%.

PR #39, opened 2026-07-23
feat: match dvd filesystem

cypressru:
Adds the DVD filesystem implementation, including FST initialization, case-insensitive path traversal, file opening and closing, entry-to-path conversion, current-directory lookup, asynchronous and synchronous reads, and transferred-size reporting.

Every linked function was verified byte-for-byte in objdiff, and the object passed the fresh-build, section-size, Matching-link, DOL SHA-1, and all 17 REL SHA-1 gates.


Jovinull:
Merged. Two conflicts worth flagging:

1. `splits.txt` — this PR moves `__mem.c` and `__start.c` ahead of `dvdfs.c`, while main had them sitting between `dvdidutils.c` and `dvdFatal.c`. Git merged that as two duplicate copies of both blocks. Kept a single pair in your position and restored `dvdidutils.c`/`dvdFatal.c`/`fstload.c` after `dvderror.c`.
2. `dvdfs.c` called `fn_801D2C4C` through a `#define OSPanic` shim. #34 named that symbol `OSPanic`, so the link failed. Declared it in `include/dolphin/os.h` and dropped the shim — same as `OSGetStackPointer` in #34.

Rebuilt clean: SHA-1 gate 18/18, dvdfs.c 12/12 at 100%, all neighbouring dvd objects still 100%.

PR #42, opened 2026-07-24
feat: match game heap

cypressru:
Adds the game’s heap initialization, allocation, reallocation, clearing, release, free-list splitting, and block coalescing routines.

Matching requires exception tables to be enabled while propagation and instruction scheduling remain disabled. The realloc and free routines additionally disable peephole optimization locally, preserving their original entry comparisons and free-list selection branches.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -396,7 +396,11 @@ config.libs = [
         "cflags": cflags_base,
         "progress_category": "game",
         "objects": [
-            Object(NonMatching, "game/heap.c", extra_cflags=["-opt noschedule"]),
+            Object(
+                Matching,
+                "game/heap.c",
+                extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopropagation"],
+            ),
             Object(NonMatching, "game/main.c", extra_cflags=["-opt noschedule"]),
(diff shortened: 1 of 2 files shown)


Jovinull:
Merged. This branched from before heap.c moved on main, so it conflicted with three commits there (`fce99a3`, `f16c1aa`, `3da75b8` — `fn_80012654`, `fn_8001234C`, `fn_80012A94` at 100%).

Resolved by taking your heap.c wholesale rather than merging hunk by hunk: yours is 11/11 where main was 8/11, and it gets there with `-Cpp_exceptions on` and `-opt noschedule,nopropagation`, which main did not have. No point splicing a partial version into a complete one.

Worth noting we found the same thing independently — the init calling `OSSetArenaLo` a second time with the arena top, so the heap claims the whole arena.

Verified after the merge: SHA-1 gate 18/18, game/heap.c 11/11 at 100% and now linked, clang-format clean.

Heads up so we do not collide again: I am on `dolphin/os/OSThread.c` now, a new TU carved out of the gap between OSSync and OSTime (21 functions, 10 matching). Not touching game/ any more.

PR #43, opened 2026-07-24
feat: match game main

cypressru:
Adds the game’s startup, shutdown, main-loop initialization, argument dispatch, video reset, and state accessor routines.

One matching-sensitive detail is that the two tick words copied into the dispatcher arguments must use a single eight-byte structure assignment. Two independent assignments are semantically equivalent, but change GC/1.3.2 volatile register allocation. The unit also requires exception tables to be enabled and instruction scheduling disabled.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -399,7 +399,11 @@ config.libs = [
         "progress_category": "game",
         "objects": [
             Object(NonMatching, "game/heap.c", extra_cflags=["-opt noschedule"]),
-            Object(NonMatching, "game/main.c", extra_cflags=["-opt noschedule"]),
+            Object(
+                Matching,
+                "game/main.c",
+                extra_cflags=["-Cpp_exceptions on", "-opt noschedule"],
+            ),
         ],
     },
     {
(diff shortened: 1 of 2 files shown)


Jovinull:
Merged. Conflict was configure.py only — this branched before heap.c became Matching, so its side still carried the old entry; resolution keeps both.

The eight-byte structure assignment is the answer to the six instructions main was stuck on. I had ground through ~578 variants on that function (padding, local orderings, declared widths, pointer/temporary/chained forms of that block) and never tried a struct copy — two word assignments are semantically identical and I had no reason to think the copy width mattered. Good find.

Verified after merge: SHA-1 18/18, game/main.c 7/7 and game/heap.c 11/11, so game/ is complete. clang-format clean.

Still on `dolphin/os/OSThread.c` here, 10/21. One thing from it that may be useful to you: that unit needed `-inline noauto` plus the `inline` keyword on the three helpers the original actually expands. Under `-inline auto` the compiler folded `OSWakeupThread` and `__OSGetEffectivePriority` into their callers where the original emits `bl`, worth forty to fifty points on two functions.

PR #44, opened 2026-07-24
Pr osthread complete

cypressru:
Adds the Dolphin OS thread scheduler, thread creation and termination, priority inheritance, suspension and resumption, sleep and wake queues, cancellation, and stack initialization.

Every function was verified byte-for-byte in objdiff, and the object passed the fresh-build, section-size, Matching-link, and DOL and REL SHA-1 gates.

One matching-sensitive detail is that the active-thread queue must be represented through its low-memory pointer array. Advancing the tail pointer in C reproduces MetroWerks’ update-form load and exact register allocation. The unit also requires -inline noauto, with the SDK’s selectively inlined priority and rescheduling operations expressed explicitly in C.


Jovinull:
Merged. This branched from before OSThread moved on main, so it conflicted with `1f95f03` (10/21 there). Took your OSThread.c and symbol names wholesale rather than splicing — yours is 21/21, no point merging a partial into a complete one. Same call I made on #42.

Two notes on the resolution:
- We both landed on `-inline noauto` independently. Kept the comment from main explaining which helpers the original expands and which it calls, over the bare entry.
- Took your wider `.bss` bound (to 0x803F2DB0, the end of the gap) over mine, which stopped after RunQueue.

Verified after merge: SHA-1 18/18, OSThread.c 21/21 at 100% and linked, clang-format clean. SDK is now 97.5% and `dbcomm` is the only unit left below 100% in the DOL.

Separately, main now has the first REL translation unit written — `autosaveD/table.c`, four functions at the tail of that module's .text, 4/4 with the REL still reproducing. Worth knowing before you start on modules: a hand written TU is not force-exported the way dtk's own objects are, so any function in it that is only called from outside the module gets dead-stripped and the REL comes out short. Fix is `force_active` on the module in config.yml — there is a worked example in there now.

PR #45, opened 2026-07-24
feat: match dolphin dbcomm

cypressru:
Adds the OdemuExi2 debugger communication driver, including EXI transfers, mailbox reads and writes, debugger interrupt handling, payload polling, and block transfers.

Every function was verified byte-for-byte in objdiff, and the object passed the fresh-build, section-size, Matching-link, and DOL and REL SHA-1 gates.

One matching-sensitive detail is that DBGEXIImm must retain the original SDK’s C long parameter and local types rather than the project’s int-based s32 and u32 aliases. Although both are 32-bit on this target, the distinction changes GC/1.2.5n’s loop optimization. The object also uses the OdemuExi2 runtime flags, including deferred inlining, signed chars, disabled common symbols, and load/store-multiple generation.


Jovinull:
Merged, clean — no conflicts, and it did not touch anything I had in flight.

Tested standalone before merging as usual: 12/12 functions and the data all at 100%, SHA-1 18/18. Re-verified after the merge with everything together — same result, and OSThread.c, autosaveD/table.c and the game units are untouched at 100%.

This clears the last unit in the DOL that was still below 100%. Every carved DOL unit now matches except `dolphin/si/SIBios.c`, which I carved a few commits ago (`188e69f`) and is at 7/22 — so that is the only open front on the DOL side if you want to pick some of it up. The remaining fifteen are mostly the transfer and type-detection paths: `__SITransfer` at 0x801F95C8, `SIInterruptHandler`, `GetTypeCallback`, `SIGetType`/`SIGetTypeAsync`.

Two things from it that will save you time there: the alarm handler at 0x801F9B80 divides a pointer difference by 40 to recover a channel index, which is sizeof(OSAlarm) and pins the 160-byte array next to Packet as `OSAlarm Alarm[4]`. And `SISetXY` only matches written as two statements (`reg = x << 16;` then `reg |= y << 8;`) — as a single or-expression the compiler uses a scratch register instead of accumulating into the one that survives the call.

PR #47, opened 2026-07-24
feat: match autosave selector

cypressru:
Adds the autosave menu selector’s left/right input handling, index adjustment, bounds update, and selected-entry retrieval.

The function was verified byte-for-byte in objdiff, and the object passed the fresh-build, section-size, Matching-link, and DOL and all 17 REL SHA-1 gates.

One matching-sensitive detail is that this translation unit uses the same noschedule, nopeephole optimization settings as the adjacent autosave table unit. The selector’s entry array also occupies the first 0x84 bytes of its containing object, allowing the current index at offset 0x84 to produce the original indexed load exactly.


cypressru:

PR #48, opened 2026-07-24
feat: match autosave dual selector

cypressru:
Adds the autosave menu’s two-controller selector handling, including left and right input processing, index bounds updates, and selected-entry retrieval.

One matching-sensitive detail is that this entry point is not referenced internally by the REL. It therefore must be marked force_active; otherwise MetroWerks removes the compiled function and shortens the autosave module by 0xFC bytes. The translation unit also uses the adjacent autosave code’s noschedule,nopeephole optimization settings.


cypressru:

PR #49, opened 2026-07-24
feat: match autosave grid selector

cypressru:
Adds the autosave menu’s four-entry grid navigation, including directional D-pad transitions, controller-port normalization, index validation, and selected-entry retrieval.

One matching-sensitive detail is that the four menu entries form a two-by-two grid rather than a linear list. Preserving the original switch case ordering produces MetroWerks’ exact comparison tree and branch layout for the 0x1BC-byte function. The entry point must also remain force_active because it is invoked externally rather than referenced inside the autosave REL.


cypressru:

PR #50, opened 2026-07-24
feat: match autosave horizontal selector

cypressru:
Adds the autosave menu’s per-controller horizontal selector handling, including left and right D-pad input, index bounds updates, and selected-entry retrieval.

One matching-sensitive detail is that this function uses the D-pad masks 1 and 2 for horizontal navigation, while the neighboring selector routines use 4 and 8 for vertical navigation. The entry point must also remain force_active because nothing inside the autosave REL references it; otherwise MetroWerks removes the complete 0xB8-byte function


cypressru:

PR #51, opened 2026-07-24
feat: match autosave variable grid selector

cypressru:
Adds the autosave selector helper that enables index wrapping.

The entry point is compiled in the autosave module’s evidenced C++ language mode with C linkage. It remains force_active because it is externally invoked rather than referenced inside the REL.


cypressru:
Superseded by #99, which contains this function in the complete autosave selector-core translation unit. Closing this fragment to prevent duplicate address/object definitions when the pending autosave PR set is merged together.

PR #52, opened 2026-07-24
feat: match autosave state initializer

cypressru:
Adds initialization of the autosave window state, clearing its runtime fields, copying the saved tick pair, and establishing its enabled defaults.

One matching-sensitive detail is that the five floating-point fields must use the autosave module’s existing constant-zero symbol. Writing literal 0.0f values produces the same instructions but assigns the relocation to a new translation-unit constant, while a non-const external causes MetroWerks to reload it for each field. The two final flags also remain a chained assignment to preserve the original reverse store order.


cypressru:

PR #53, opened 2026-07-24
feat: match autosave window no-op callback

cypressru:
Adds the autosave window’s empty virtual callback used by its method table.

The callback intentionally has an empty C body, producing the original four-byte blr. It is compiled in the autosave module’s evidenced C++ language mode with C linkage and is retained naturally through its method-table reference.


cypressru:

PR #54, opened 2026-07-24
feat: match autosave selector index setter

cypressru:
Adds the autosave selector index setter, accepting only the valid nonzero selector indices used by the autosave menu.

One matching-sensitive detail is that the lower and upper bounds remain separate early-return checks. This reproduces the original blelr and bgelr instruction sequence exactly. The entry point is compiled in the autosave module’s evidenced C++ language mode with C linkage.


cypressru:

PR #55, opened 2026-07-24
feat: match autosave wrap enable

cypressru:
Adds the autosave selector helper that enables index wrapping.

The entry point is compiled in the autosave module’s evidenced C++ language mode with C linkage. It remains force_active because it is externally invoked rather than referenced inside the REL.


cypressru:

PR #56, opened 2026-07-24
feat: match autosave select value

cypressru:
Adds autosave selector lookup by entry value, updating the current index and applying the shared bounds helper when a matching entry is found.

One matching-sensitive detail is that the scan must use an equality-terminated loop (i != count). The equivalent i < count form changes MetroWerks’ empty-list guard from the original beq to ble. The entry point is compiled in the autosave module’s evidenced C++ language mode with C linkage and remains force_active because it is externally invoked rather than referenced inside the REL.


cypressru:

PR #57, opened 2026-07-24
feat: match autosave selector bounds

cypressru:
Adds the autosave selector’s shared bounds handling, including optional index wrapping and non- wrapping endpoint clamping.

One matching-sensitive detail is that the upper-bound checks must be expressed as count  > index followed by an early return. The equivalent index >= count condition reverses MetroWerks’ compare operands and produces bltlr instead of the original bgtlr. The helper is compiled in the autosave module’s evidenced C++ language mode while retaining its original C linkage.


cypressru:

PR #58, opened 2026-07-24
feat: match autosave shutdown wrapper

cypressru:
Adds the autosave shutdown wrapper, releasing the module’s loaded window resources before destroying its shared runtime resource.

Both operations remain direct calls in their original order. The function and its callees use C linkage while the translation unit retains the autosave module’s evidenced C++ language mode.


cypressru:

PR #60, opened 2026-07-24
feat: match autosave window constructor

cypressru:
Adds construction of the autosave window, including base and selector initialization, state copying, widget setup, resource-backed selector entries, mode defaults, and the initial selection derived from saved resource state.

The function was verified byte-for-byte in objdiff, with its 636-byte .text and owned 8-byte .rodata
both matching 100%. The object also passed the forced fresh-build, section-size, Matching-link, DOL SHA-1, and all 17 REL SHA-1 gates.

One matching-sensitive detail is that the constructor owns the unsigned-integer conversion constant, while the adjacent zero-float constant remains owned by the neighboring translation unit because its remaining assembly still references the original global symbol. Splitting the constants at that boundary preserves both MetroWerks code generation and correct linker symbol resolution.

### DIFF
--- a/config/G9SE8P/autosaveD/splits.txt
+++ b/config/G9SE8P/autosaveD/splits.txt
@@ -6,5 +6,9 @@ Sections:
 	.data       type:data align:8
 	.bss        type:bss align:8
 
+autosaveD/window_constructor.c:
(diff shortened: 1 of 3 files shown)


cypressru:

PR #61, opened 2026-07-24
feat: match autosave state copy

cypressru:
Adds the autosave state copy helper, copying the complete 68-byte state record including its four
paired subobjects.

One matching-sensitive detail is that each paired subobject is copied through an inline helper with an ordered volatile source. This preserves the original first-word then second-word loads and MetroWerks’ corresponding r5/r0 allocation; an ordinary structure assignment instead becomes a compact copy loop or reverses the temporary registers.


cypressru:

PR #62, opened 2026-07-24
feat: match autosave window factory

cypressru:
Adds the autosave window factory, allocating the complete window object and invoking its constructor with the supplied parent and configuration when allocation succeeds.

The allocation result is returned unchanged when null and replaced with the constructor result
otherwise, reproducing the original C++ allocation pattern and register flow.


cypressru:

PR #63, opened 2026-07-24
feat: match autosave resource cleanup

cypressru:
Adds cleanup of the autosave window resources, clearing the five cached asset pointers and releasing the mounted archive when present.

One matching-sensitive detail is that the cached assets must be cleared through a fixed five-
iteration pointer loop. MetroWerks fully unrolls that loop into the original alternating stw and addi
 > sequence; writing five equivalent assignments directly instead produces base-plus-offset stores.


cypressru:

PR #64, opened 2026-07-24
feat: match autosave resource loader

cypressru:
Adds loading of the autosave window archive, including resource construction, temporary archive-buffer allocation, mounting, lookup of the five window assets, and cleanup of the temporary resources.

One matching-sensitive detail is that allocation and conditional resource construction remain
isolated in an inline helper. The helper disappears completely after inlining, but preserves
MetroWerks’ original live-range ordering: the resource stays in r30, while the generated asset-name and output iterators use r29 and r28.


cypressru:

PR #65, opened 2026-07-24
feat: match autosave window state getter

cypressru:
Adds the autosave window state accessor, returning the current state stored at offset 0x6C.

The function’s 8-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The explicit window layout preserves the original field offset while keeping the implementation
entirely in C.


cypressru:

PR #66, opened 2026-07-24
feat: match autosave window configuration

cypressru:
Adds autosave window configuration, constructing the local timing and color parameters before
forwarding them to the window setup routine.

The function’s 96-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

One matching-sensitive detail is that the function must be listed as force-active. Nothing in the
remaining module directly references its compiled symbol, so MetroWerks otherwise dead-strips it and shifts the resulting REL despite the function itself matching exactly.


cypressru:

PR #67, opened 2026-07-24
feat: match autosave window destructor

cypressru:
Adds destruction of the autosave window, restoring its method table, destroying the embedded selector and base window, and conditionally releasing the allocation according to the destructor flags.

The function’s 120-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

One matching-sensitive detail is that the deletion flag must be narrowed to a signed 16-bit value
before comparison. This reproduces the original extsh instruction used by MetroWerks destructor dispatch while retaining an ordinary C++ implementation.


cypressru:

PR #68, opened 2026-07-24
feat: match autosave widget rendering

cypressru:
Adds rendering of an autosave widget, including its initial setup, selected or unselected shared resource binding, geometry rendering, and finalization.

The function’s 156-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

One matching-sensitive detail is preserving all four incoming arguments across the initial setup
call. The straightforward staged implementation causes MetroWerks to retain them in r28 through r31, reproducing the original register allocation and call sequence.


cypressru:

PR #70, opened 2026-07-24
feat: match autosave widget overlay

cypressru:
Adds rendering of the autosave widget overlay, including setup, binding of its shared overlay
resource, geometry submission, and finalization.

The function’s 112-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

One matching-sensitive detail is keeping the context, position, and size arguments live across the setup and resource-binding calls. This reproduces MetroWerks’ original r29 through r31 allocation and exact call sequence.


cypressru:

PR #73, opened 2026-07-24
feat: match autosave task cleanup

cypressru:
Adds the autosave task destruction marker, setting the low bit of the task’s 16-bit flags field.

The function’s 20-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

Keeping the flags field as u16 is matching-sensitive: the compound assignment reproduces MetroWerks’ original halfword load, bitwise OR, 16-bit truncation, and halfword store sequence.

### DIFF
--- a/config/G9SE8P/autosaveD/splits.txt
+++ b/config/G9SE8P/autosaveD/splits.txt
@@ -6,5 +6,8 @@ Sections:
 	.data       type:data align:8
 	.bss        type:bss align:8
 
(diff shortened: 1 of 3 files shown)


cypressru:
Superseded by #100, which contains this function in the complete autosave task-runtime translation unit. Closing this fragment to prevent duplicate address/object definitions when the pending autosave PR set is merged together.

PR #76, opened 2026-07-24
feat: match autosave heap allocation wrapper

cypressru:
Adds the autosave module's heap-allocation wrapper, allocating the requested byte count from the game's global heap and returning the resulting pointer.

The function's 48-byte .text section was verified byte-for-byte in objdiff at 100%. The object also passed the forced fresh-build, section-size, Matching-link, DOL SHA-1, and all 17 REL SHA-1 gates.

The wrapper and underlying allocator both retain pointer return types. This documents the behavior used by callers while preserving the original heap-pointer load, size argument handoff, and returned r3 value.


cypressru:
Caller analysis identified this function as the autosave module's heap-allocation wrapper rather than task unregistration. I updated the PR branch in place: the source is now heap_alloc.c, the wrapper has the correct pointer return type and byte-count parameter, and the code remains 100% byte-for-byte with all 18 SHA-1 outputs passing.

cypressru:

PR #78, opened 2026-07-24
feat: match autosave module shutdown

cypressru:
Adds autosave module shutdown, unregistering the module’s global state before shutting down its supporting subsystem and autosave resources.

The function’s 48-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The module state remains declared as the original 24-byte object at .data+0xA8, preserving the
module-relative address relocation and exact three-call teardown order.


cypressru:

PR #79, opened 2026-07-24
feat: match autosave task callbacks

cypressru:
Adds five empty autosave task callback entry points used by the task lifecycle interface.

The TU’s complete 20-byte .text section—five four-byte functions—was verified byte-for-byte in
objdiff at 100%. The object also passed the forced fresh-build, section-size, Matching-link, DOL SHA-1, and all 17 REL SHA-1 gates.

The callbacks remain separate exported functions despite sharing empty implementations, preserving their original symbol boundaries and ordering.


cypressru:

PR #81, opened 2026-07-24
feat: match autosave task callback cleanup

cypressru:
Adds conditional cleanup of the autosave task callback object stored at offset 0x2C.

The callback field is intentionally left unchanged after the cleanup call, matching the original
lifecycle behavior and exact MetroWerks instruction sequence.


cypressru:
Superseded by the complete pr-autosave-task-object translation-unit branch: https://github.com/Jovinull/sonicheroes/compare/main...cypressru:pr-autosave-task-object?expand=1. Closing this fragment to prevent duplicate address/object definitions when the pending autosave PR set is merged.

PR #82, opened 2026-07-24
feat: match autosave task destructor

cypressru:
Adds the autosave task deleting destructor. It restores the task class table, performs subsystem and base-object teardown, and conditionally releases the allocation through the module heap.

The function’s 116-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The deletion flags remain a signed 16-bit parameter, reproducing MetroWerks’ explicit sign extension before the conditional heap-free call. The destructor returns the original task pointer on every path.

### DIFF
--- a/config/G9SE8P/autosaveD/splits.txt
+++ b/config/G9SE8P/autosaveD/splits.txt
@@ -6,5 +6,8 @@ Sections:
 	.data       type:data align:8
 	.bss        type:bss align:8
 
+autosaveD/task_destructor.c:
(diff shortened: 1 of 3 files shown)


cypressru:
Superseded by the complete pr-autosave-task-object translation-unit branch: https://github.com/Jovinull/sonicheroes/compare/main...cypressru:pr-autosave-task-object?expand=1. Closing this fragment to prevent duplicate address/object definitions when the pending autosave PR set is merged.

PR #83, opened 2026-07-24
feat: match autosave task creation

cypressru:
Adds autosave task creation, validating the input context and subsystem state before allocating a 56- byte task object and running its constructor.

The function’s 104-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The function remains a void creation callback: it deliberately ignores the constructor’s returned
pointer, matching the original control flow and register behavior. Allocation is skipped for a null
context, while an inactive subsystem reports type 2.


cypressru:

PR #84, opened 2026-07-24
feat: match autosave task system cleanup

cypressru:
Adds autosave task-system cleanup, conditionally tearing down the active rendering state and clearing its associated resource descriptor.

The function’s 64-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

Cleanup is skipped when the task subsystem is inactive, matching the original callback’s guarded control flow.


cypressru:

PR #86, opened 2026-07-24
feat: match autosave controller status

cypressru:
Adds the autosave controller-status adapter, mapping autosave states to the appropriate controller slot and returning the underlying input status unchanged.

The function’s 116-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

State 1 selects controller slot 0, state 2 selects slot 1, and all other states use the default
controller path. State 0 remains an explicit switch case because that source detail preserves the
original MetroWerks comparison tree.


cypressru:

PR #87, opened 2026-07-24
feat: match autosave parameter synchronization

cypressru:
Adds autosave parameter synchronization, copying ten fields from the autosave state object into their corresponding engine parameter slots and clearing the final control flag.

The function’s 172-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The parameter calls intentionally preserve the original non-sequential ID order while reading the
source values contiguously from offsets 0x810 through 0x834.


cypressru:

PR #88, opened 2026-07-24
feat: match autosave parameter snapshot

cypressru:
Adds autosave parameter snapshotting, preserving ten engine parameters in the autosave state object before installing the temporary defaults required by the autosave screen.

The function’s 280-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The snapshot and replacement operations intentionally use the same non-sequential parameter ID order as the corresponding restoration routine, allowing the original settings to be restored when the autosave screen exits.


cypressru:

PR #89, opened 2026-07-24
feat: match autosave screen rendering

cypressru:
Adds the mode-selecting autosave render wrapper, applying one of two saved render-enable values before drawing through the alternate autosave presentation path and restoring the previous engine parameters afterward.

The function’s 156-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

A nonzero mode selects the first saved render value, while zero selects the second. The surrounding snapshot and restoration calls ensure the temporary render configuration remains scoped to this draw operation.

### DIFF
--- a/config/G9SE8P/autosaveD/splits.txt
+++ b/config/G9SE8P/autosaveD/splits.txt
@@ -6,5 +6,8 @@ Sections:
 	.data       type:data align:8
 	.bss        type:bss align:8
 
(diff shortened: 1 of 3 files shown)


cypressru:
Duplicate address range of #70 (0x32F8-0x3368). Closing the later re-extraction to keep one owner for this code.

PR #90, opened 2026-07-24
feat: match autosave result getter

cypressru:
Adds the autosave result getter, returning the completion/result field stored at offset 0x6C in the
autosave state object.

The function’s 8-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The field is updated by the autosave state logic and queried by several higher-level controllers to
determine the current completion outcome.


cypressru:
Duplicate address range of #65 (0x3404-0x340C). Closing the later re-extraction to keep one owner for this code.

PR #92, opened 2026-07-24
feat: match autosave default rendering

cypressru:
Adds the autosave default-render wrapper, constructing a default placement block from the global display state and forwarding it with a black color at 0xA0 alpha to the lower-level renderer.

The function’s 96-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The packet fields use explicit assignments rather than aggregate initialization because aggregate initialization introduces hidden constant copies and changes MetroWerks code generation. The function is also marked force-active because it has no internal callers and would otherwise be dead-stripped after separation from the original monolithic object.

### DIFF
--- a/config/G9SE8P/autosaveD/splits.txt
+++ b/config/G9SE8P/autosaveD/splits.txt
@@ -6,5 +6,8 @@ Sections:
 	.data       type:data align:8
 	.bss        type:bss align:8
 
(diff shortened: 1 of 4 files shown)


cypressru:
Duplicate address range of #66 (0x340C-0x346C). Closing the later re-extraction to keep one owner for this code.

PR #93, opened 2026-07-24
feat: match autosave render state initialization

cypressru:
Adds autosave render-state initialization, initializing the shared engine render context before resetting the autosave render slots and releasing any stale render resource.

The initialization order is preserved from the original: the shared context at lbl_803E8150 is
initialized first, followed by autosave-specific render-state cleanup.


cypressru:
Duplicate address range of #59 (0x3FAC-0x3FD8). Closing the later re-extraction to keep one owner for this code.

PR #94, opened 2026-07-24
feat: match autosave render state shutdown

cypressru:
Adds autosave render-state shutdown, releasing the autosave render resources before shutting down the shared engine render context.

The teardown order is preserved from the original: autosave-specific resources are released first, followed by the shared context at lbl_803E8150.


cypressru:
Duplicate address range of #58 (0x3FD8-0x4004). Closing the later re-extraction to keep one owner for this code.

PR #95, opened 2026-07-24
feat: match autosave render parameter initialization

cypressru:
Adds autosave render-parameter initialization, clearing the parameter block, initializing its
floating-point fields, copying the shared timing values, and enabling its two final state flags.

The function’s 108-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

The shared timing values are copied directly from lbl_8029BB80, preserving the original load and store ordering.


cypressru:
Duplicate address range of #52 (0x4004-0x4070). Closing the later re-extraction to keep one owner for this code.

PR #97, opened 2026-07-24
feat: match autosave render callback no-op

cypressru:
Adds the autosave render callback no-op referenced by the module’s callback table.

The function’s four-byte .text section was verified byte-for-byte in objdiff at 100%. The object also


cypressru:
Duplicate address range of #53 (0x4070-0x4074). Closing the later re-extraction to keep one owner for this code.

PR #98, opened 2026-07-24
feat: match autosave render mode setter

cypressru:
Adds the autosave render-mode setter, accepting only the two valid nonzero modes before updating the render state.

The function’s 24-byte .text section was verified byte-for-byte in objdiff at 100%. The object also

Signed range checks are retained so the compiler emits the original early-return blelr and bgelr
sequence.


cypressru:
Superseded by #99, which reconstructs the complete four-function selector_core translation unit and includes this setter.

PR #99, opened 2026-07-24
feat: match autosave selector core

cypressru:
Adds the autosave selector core, including validated index assignment, wrap enabling, value-to-index lookup, and index normalization with optional wrapping.

All four functions and the complete 236-byte .text section were verified byte-for-byte in objdiff at
100%. The object also passed the forced fresh-build, section-size, Matching-link, DOL SHA-1, and all 17 REL SHA-1 gates.

The wrap setter and value lookup remain force-active because they have no static callers after
extraction but are externally reachable in the original module. Without retaining the lookup
function, the rebuilt REL is exactly 88 bytes short.


cypressru:

PR #114, opened 2026-07-26
feat: match MSL multibyte strings

cypressru:
Adds the MSL multibyte and wide-character conversion translation unit, including UTF-8 encoding, decoding, validation, and string conversion.

Every function was verified byte-for-byte in objdiff, and the object passed the fresh-build, section-size, Matching-link, and DOL/all 17 REL SHA-1 gates.

One matching-sensitive detail is that wcstombs returns and stores its converted-byte count as C unsigned long, while its limit parameter must remain unsigned int; although both are 32-bit on this target, the distinction changes GC/1.3 register allocation. wctomb also keeps its UTF-8 conversion directly in the public function, because routing it through an inline helper changes the null-return register and duplicates the pooled four-byte prefix table.


Jovinull (approved):
Validado localmente sobre o main atual (rebase sobre 237fcb2): build reproduz os 18/18 SHA-1 e a unidade fecha em 100% (880/880 bytes, 5/5 funcoes). clang-format limpo. Os quatro simbolos renomeados batem com os tamanhos que o dtk ja tinha.

PR #116, opened 2026-07-26
AutoSaveD rework

ThePlayerRolo:
a lot of progress to make this match the ps2 symbols more

still a draft


Jovinull:
Before we formalize the repository-wide C/C++ policy, could you please document the exact evidence behind these two claims from Discussion #117?

1. That main and heap are the only C-linkage units in the main game, with representative PS2 mangled/unmangled symbol names or a public symbol index/reference.
2. That the main REL sources used -inline deferred, with representative units and the observed source/emission order that supports it.

Please share only legally redistributable public references or small textual excerpts limited to symbol names/order. Do not upload or link game/prototype binaries, disc images, extracted assembly, or leaked source. We will treat the GameCube binary as the codegen authority and PS2 evidence as naming/linkage guidance. We also will not touch or duplicate the AutoSaveD work in this draft while auditing the policy.

ThePlayerRolo:
> Before we formalize the repository-wide C/C++ policy, could you please document the exact evidence behind these two claims from Discussion #117?
> 
> 1. That main and heap are the only C-linkage units in the main game, with representative PS2 mangled/unmangled symbol names or a public symbol index/reference.

I don't think they are the **only** ones but they are some. This is because according to ps2 symbols they don't have c++ mangling (suggesting they either had extern c or just used c) 

for example at 00303e40 in the ps2 binary there is  the symbol MAlloc_BW (note how this symbol has no mangling?)

This symbol is used in ADV_WINDOW::InitializeCore with the code matching the gc version

`void* buffer = MAlloc_BW(0x300000);`

In the gc version MAlloc_BW is in 0x80012994, found in heap.c

FYI I also used https://github.com/chaoticgd/ccc to extract the symbol data 

> 2. That the main REL sources used -inline deferred, with representative units and the observed source/emission order that supports it.
> 

First of all, ADV_WINDOW's vtable in ADV_WINDOW.cpp needs to be placed before a string, which doesn't happen to my knowledge unless inline differed is used. 
There is also the fact that certain functions in that TU would get inlined if Not for deferred. (InitializeCore in Initialize) 

As well as this, the prolog, epiloug, and unresolved functions are flipped from where they normally are in most rels, a characteristic of -inline deferred

Jovinull:
Thanks — the `MAlloc_BW` correlation is useful. One important distinction from the audit: an unmangled symbol proves C linkage, not necessarily a C translation unit. In the same PS2 memory block, `psAllocateHeapFromKernel__Fv`, `private_calloc__FUiUi`, `private_realloc__FPvUi`, `private_free__FPv`, and `private_malloc__FUi` are C++-mangled, while `psGetMemoryFunctions`, `Free_BW`, and `MAlloc_BW` retain the public C ABI.

The GameCube tests now separate the two cases:

- `heap`: C++ preserves all function code, and C++ declaration-order `.sbss` reproduces the target directly; the previous C form needed reversed declarations. It is now `heap.cpp` with explicit `extern "C"` only on the public boundary, and all 18 artifact hashes pass.
- `main`: C remains justified by the GameCube target. A direct C++ compile grows `main` by 8 bytes; C++ plus the correct C ABI and typed equivalents restores the size but still changes six instructions. The PS2 platform file being C++ does not override that target-specific result.

For `-inline deferred`, the current draft is not yet sufficient evidence for a blanket AutoSaveD setting. It applies `-inline noauto, deferred` to every fragment (including `prolog.c`), and `ADV_MENU.cpp` currently has two object-level `-inline` flags. Please record, per reconstructed logical TU, the exact command, the before/after objdiff, the emitted function/data order, and which specific mismatch is fixed by deferred. Vtable/string placement and `_prolog`/`_epilog` order are useful leads, but can also reflect declaration, section, or link order. The language-policy checker intentionally starts with an empty deferred allowlist and will reject duplicate or undocumented overrides.

One provenance point before this draft leaves draft status: `include/rwsdk/README.md` says the headers were ripped from BFBB, and the blobs match BFBB paths. BFBB declares CC0, but please pin the exact upstream commit/license and confirm these are redistributable reconstructed declarations rather than copied proprietary SDK material. If that cannot be established, keep only the minimal clean-room declarations required here. No binaries, SDK archives, extracted source, or symbol dumps are needed.

I will continue to leave all AutoSaveD source/splits untouched while this draft is active.

ThePlayerRolo:
> 
> For `-inline deferred`, the current draft is not yet sufficient evidence for a blanket AutoSaveD setting. It applies `-inline noauto, deferred` to every fragment (including `prolog.c`), and `ADV_MENU.cpp` currently has two object-level `-inline` flags. Please record, per reconstructed logical TU, the exact command, the before/after objdiff, the emitted function/data order, and which specific mismatch is fixed by deferred. Vtable/string placement and `_prolog`/`_epilog` order are useful leads, but can also reflect declaration, section, or link order. The language-policy checker intentionally starts with an empty deferred allowlist and will reject duplicate or undocumented overrides.


This issue with the vtable is that as I said before, there is no way besides inline deferred to place it in that position since without it always goes to the end of the TU

Also the double ADV_MENU.cpp inline deffered thing is a remnant from older stuff

> One provenance point before this draft leaves draft status: `include/rwsdk/README.md` says the headers were ripped from BFBB, and the blobs match BFBB paths. BFBB declares CC0, but please pin the exact upstream commit/license and confirm these are redistributable reconstructed declarations rather than copied proprietary SDK material. If that cannot be established, keep only the minimal clean-room declarations required here. No binaries, SDK archives, extracted source, or symbol dumps are needed.
> 

It seems that BFBB just uses the publicly available ps2 dwarf info for types

Jovinull:
Audit update from `audit/cpp-language-migration`:

- The four current AutoSaveD C-mode fragments are now classified as protected C++ migration debt from direct `ADV_MENU`/`ADV_WINDOW`/`sADV_WINDOW_PARAM` symbols. We are still leaving their source and splits entirely to this draft.
- The central checker now blocks new game-owned `.c`, unreviewed `-inline deferred`, and duplicate object-level `-inline` overrides. Current audit result is 0 unresolved language decisions, 5 protected C++/C-mode debts, 1 reviewed vendor compiler-mode exception, and 0 approved deferred TUs.

Before integration, please update this draft in these specific ways:

1. Drop the `game/main.c` and `game/heap.c` changes. Target-side tests keep main as C; heap is already migrated to `heap.cpp` with C ABI only on the public allocator boundary, preserving all 18 hashes.
2. Remove the stale duplicate `ADV_MENU.cpp` inline override and do not apply deferred to every AutoSaveD fragment/prolog. If the reconstructed logical `ADV_WINDOW` TU requires deferred, one effective flag on that TU is fine once the PR records the exact without/with-deferred commands, objdiff, and emitted vtable/string/function order showing the target mismatch it fixes.
3. Pin the BFBB source commit and license for every reused header, and confirm the declarations are redistributable reconstruction from public symbol/type information. Otherwise reduce them to minimal project-owned declarations.
4. Rebase after the policy branch is ready, then run the language checker, full build and 18-file hash check.

The claim that deferred is the only way to place this vtable is a useful hypothesis; the requested two-command/object comparison is what will turn it into per-TU evidence. Ping me when the draft updates and I will re-audit it without duplicating the AutoSaveD work.

ThePlayerRolo:
>     2. Remove the stale duplicate `ADV_MENU.cpp` inline override and do not apply deferred to every AutoSaveD fragment/prolog. If the reconstructed logical `ADV_WINDOW` TU requires deferred, one effective flag on that TU is fine once the PR records the exact without/with-deferred commands, objdiff, and emitted vtable/string/function order showing the target mismatch it fixes.
> 
It seems ADV_MENU also requires inline deferred as otherwise due to the placement of ADV_MENU::Commit() everything else inlines it. I don't think we have split information though maybe they are split? However I think it makes more sense that they aren't since theres no data or anything to prove otherwise
If you want I could also apply pragma dont_inline (though that feels more fake match than just doing inlined defered)

Jovinull:
That `Commit()` observation can become valid flag evidence, but the current combined flag does not yet tell us **which part fixes which mismatch**. `-inline noauto, deferred` changes two variables at once: `noauto` can prevent `Commit()` from being folded, while `deferred` can change function/data emission order.

Please test each reconstructed logical TU with this small matrix, using the same source and all other flags unchanged:

1. inherited `-inline auto` only;
2. one object override `-inline noauto`;
3. one object override `-inline deferred` (with the inherited auto still present);
4. one object override `-inline noauto,deferred`.

For `ADV_MENU.cpp`, record whether `Commit()` remains an out-of-line target function, whether each target call remains `bl Commit`, function order, and objdiff for each candidate. For `ADV_WINDOW.cpp`, also record the vtable/string offsets and relocations. Then we can distinguish:

- only `noauto` is needed to preserve calls;
- only `deferred` is needed for emission/data order; or
- both are independently needed.

Please do **not** add `#pragma dont_inline` just to force the result. Without source-level evidence for that pragma it is more speculative than a TU-wide compiler option. If the four-way comparison shows the original object behavior, one exact object-level override is the cleaner reconstruction and we can allowlist that TU with the recorded evidence.

The lack of private data is not evidence that files were split or unsplit. Keep the current logical TU as an explicit boundary hypothesis and record what supports it (class/source metadata, neighboring ranges, private vtable/data, extab/relocations), without presenting the boundary as proven. A per-function extab entry alone is not a TU boundary.

As a concrete example, the audit just integrated `game/state_accessor.cpp`: ordinary auto emits the reverse declaration order, while deferred emits the exact GameCube address order, so that one TU is now the sole reviewed deferred exception. The same comparison showed `dvd_status.cpp` did not need deferred, so its flag was removed. That is the evidence standard we should apply here.

The current PR commit still applies `-inline noauto,deferred` to every AutoSaveD fragment and `prolog.c`, and still has the duplicate `ADV_MENU.cpp` override. Please narrow those only after the matrix above; no new `dont_inline` pragma is requested.

Jovinull:
Independent controlled result for the current `e78d889` `ADV_MENU.cpp` candidate, using the PR proposed split (`0x4074..0x47F0`) and changing only the final object-level inline option. I extracted the target object locally; no binary, assembly, dump, or local path is being attached.

| final object override | emitted order | candidate `.text` | objdiff `.text` | result |
|---|---|---:|---:|---|
| none (inherited auto) | constructor through `SetSelect` (opposite target order) | 1992 | 94.718% | `Set` absorbs `Commit`; `Set` is 0% |
| `-inline noauto` | same wrong order | 1912 | 99.311% | 12/13 functions exact; `UpdateQUAD2` 97.273% |
| `-inline deferred` | exact target order | 2592 | 62.109% | automatic inlining expands `Set` and the update methods |
| `-inline noauto,deferred` | exact target order | 1912 | 99.311% | 12/13 functions exact; `UpdateQUAD2` 97.273% |

The target order is:

`SetSelect`, `OnLoop`, `Set`, `Commit`, `UpdateQUAD2`, `UpdateQUAD`, `UpdateLR`, `UpdateUD2`, `UpdateUD`, `Clr`, `Add`, destructor, constructor.

That is exactly what both deferred candidates emit. `noauto` is independently required for code generation: with it, `Set`, `Commit`, all update methods except the already-incomplete `UpdateQUAD2`, `Clr`, `Add`, destructor and constructor all match 100%. The combined candidate and target both have 70 relocations. The only remaining size difference is `UpdateQUAD2`: target 484 bytes, candidate 480 bytes, with 97.273% function match; this is a source reconstruction mismatch, not evidence against the flags.

So for the **proposed ADV_MENU logical TU**, one `-inline noauto,deferred` override is supported. `#pragma dont_inline` is neither needed nor preferable. Please:

1. remove the duplicate ADV_MENU override;
2. retain exactly one `-inline noauto,deferred` on `ADV_MENU.cpp`;
3. fix the remaining four-byte/objdiff mismatch in `UpdateQUAD2` before calling the object matching;
4. remove the blanket option from `prolog.c` and every other fragment until that specific TU has the same controlled evidence;
5. record separately that the GC TU boundary is still a hypothesis supported by the reconstructed class run, not established merely by this flag result.

This turns your `Commit()` observation into reproducible per-TU evidence. I will run the same isolation for `ADV_WINDOW` rather than extrapolating this result to it or to the whole REL.

Jovinull:
Independent `ADV_WINDOW.cpp` isolation found both real deferred evidence **and** structural problems that must be fixed before the TU can be reviewed as matching.

Using the PR proposed target split (`0x38F4..0x4074`) and the same four flag candidates:

| final override | candidate `.text` | objdiff `.text` | target `.data` result |
|---|---:|---:|---|
| inherited auto | 4480 | 83.271% | 48.227%, 148 bytes |
| `-inline noauto` | 3036 | 92.250% | 48.227%, 148 bytes |
| `-inline deferred` | 4576 | 78.260% | **100%, 147 bytes** |
| `-inline noauto,deferred` | 3036 | **92.490%** | **100%, 147 bytes** |

This confirms the vtable observation precisely: without deferred the candidate vtable is emitted at data offset `0x68` and the section is one byte too large; with deferred it is at target offset `0x4C`, the complete 147-byte `.data` matches 100%, and the other owned data is in the target order. `noauto` is separately needed for code: in the combined candidate, `Exec`, destructor, constructor, `Create`, `FinalizeCore`, `Finalize`, `Initialize`, the parameter constructor and `TDisp` are 100%; `InitializeCore` is 98.220%.

However the current source cannot be accepted yet:

1. `ADV_WINDOW.cpp` defines `_GetButton`, `End`, `Begin`, `DispPoly`, `DispWindow` and `DispSelect`, while `ADV_WINDOW_DISP.cpp` defines the same six functions. In the proposed `ADV_WINDOW` target object these are not definitions (`_GetButton` is an unresolved dependency); the six candidate definitions add 1260 bytes and would also create duplicate symbols at link time.
2. The target owns `sADV_WINDOW_PARAM::operator=` (140 bytes), but the implementation is commented out, so the candidate leaves it unresolved.
3. Target `.bss` is 40 bytes; candidate `.bss` is 24 bytes. The target symbol assignment gives `g_TEX` 36 bytes, while the candidate declares only five pointers (20 bytes).
4. Target `.rodata` is 16 bytes; candidate is 24 bytes and 80%.
5. With noauto, the current core definitions emit mostly in target order; deferred reverses them. Since deferred is required for `.data`, the source definitions must be arranged so deferred emits this exact target code order:
   `Exec`, destructor, constructor, parameter assignment, `Create`, `FinalizeCore`, `InitializeCore`, `Finalize`, `Initialize`, parameter constructor, `TDisp`.
   In other words, after removing the six DISP functions, source definition order should be the reverse of that list for this compiler behavior.

The size accounting is useful: combined candidate 3036 - 1260 duplicate-DISP bytes + 140 missing assignment bytes = 1916, four bytes below the 1920-byte target; that remaining four-byte gap is consistent with the still-98.220% `InitializeCore` reconstruction.

Conclusion: one `-inline noauto,deferred` override is supported for the proposed logical `ADV_WINDOW` TU because `noauto` controls calls/code expansion and `deferred` is required for exact `.data`/vtable emission. But first remove the duplicate DISP definitions, restore and match the assignment operator, fix `g_TEX`/`.bss`, rodata and `InitializeCore`, then demonstrate the final code order and 100% object result. This evidence does **not** justify deferred on `ADV_WINDOW_DISP.cpp`, prolog, or any other AutoSaveD fragment.

Jovinull:
Third controlled result: the six functions currently implemented in `ADV_WINDOW_DISP.cpp` also support `-inline noauto,deferred`, but the proposed TU is far from complete and must stay non-matching.

The source lists the implemented functions in this order:

`DispSelect`, `DispWindow`, `DispPoly`, `Begin`, `End`, `_GetButton`.

The target contains those six in the exact reverse order:

`_GetButton`, `End`, `Begin`, `DispPoly`, ...missing functions..., `DispWindow`, `DispSelect`.

Observed compiler behavior:

- inherited auto: source order; automatic expansion occurs;
- `noauto`: source order; the six bodies have target sizes/code;
- `deferred`: target order; automatic expansion makes candidate text 2336 bytes and the three display bodies mismatch badly;
- `noauto,deferred`: target order, candidate text 1260 bytes; `_GetButton`, `End`, `Begin`, `DispWindow` and `DispSelect` are 100%, and `DispPoly` is 99.858% only because its floating constant relocates to the candidate lone rodata symbol while the incomplete target TU has that constant inside its larger rodata sequence.

So both halves are again independently supported for the implemented subset: deferred fixes physical function order, noauto prevents `Begin`/`End` from being folded into display methods. No source-level `dont_inline` workaround is needed.

Completeness warning: the target split owns 5228 bytes of `.text` and 680 bytes of `.rodata`; the current candidate supplies only 1260 text bytes and 8 rodata bytes. It still lacks `RenderQuad2Herf`, `fn_2_2A18`, `fn_2_2BC8`, `RenderQuad4SKY`, `fn_2_31F0`, `fn_2_3404`, `fn_2_340C`, `Disp`, and their constants. The final source should preserve reverse target definition order so deferred emits the target order, and the flag should be formally allowlisted only after the complete TU diff and relocations pass.

This supports narrowing the draft to three candidate logical overrides (`ADV_MENU`, `ADV_WINDOW`, `ADV_WINDOW_DISP`) as each becomes complete. It still does not support applying the option to prolog or unrelated fragments.

Jovinull:
Definitive `autosaveD/prolog.c` result: remove the inline override entirely.

The current source order is `_unresolved`, `_epilog`, `_prolog`, exactly the target physical order. With only inherited auto:

- `.text`: 196/196 bytes, 100%;
- `.data`: 72/72 bytes, 100%;
- all three functions and both tables: 100%;
- raw target `.text` SHA-256 equals raw auto candidate `.text` SHA-256 (`4965484b...fee4e`).

`-inline noauto` produces the identical object as auto, so it is also unnecessary. Both deferred variants reverse physical emission to `_prolog`, `_epilog`, `_unresolved`; their raw `.text` hash is different (`ceea4e2...21fb7`). Per-symbol objdiff still reports each body as 100% because it matches symbols independently, which is why physical order and the final artifact are required checks.

Therefore the special prolog/epilog/unresolved order is evidence **against** deferred for this TU, not for it. Please leave `autosaveD/prolog.c` on inherited flags only. This also demonstrates why the blanket REL flag must be removed and each logical TU tested separately.

Jovinull:
Public-header provenance follow-up for draft commit e78d889 (this is an attribution/maintenance check, not an allegation):

The public-source comparison is reassuring.

- BFBB commit ea82f4f521ab87b035728d64ab60c08a40aac2e6 declares CC0-1.0. Fifteen of the seventeen RenderWare header files in this draft are exact BFBB blobs at that commit. rwcore.h is the same public header with the local rsglobal.h include and a final newline; rsglobal.h is a small project-specific declaration.
- SMS commit b1cfdf687911c5bcc51a5e3715f132600ab32272 also declares CC0-1.0. stdarg.h and stddef.h differ only by the source comment; float.h adds the source comment and adapts the include path; math.h adapts the include and exposes at global scope a sqrtf body already present in the same public SMS header; fdlibm.h differs only by its source comment.
- The fdlibm.h comment points to ThePlayerRolo/lm-decomp, which also declares CC0-1.0; its content is independently identical to the pinned SMS version used above.

Recommended cleanup before this leaves draft status: replace the informal word “ripped” in include/rwsdk/README.md with a neutral attribution table that pins the BFBB/SMS commits and CC0 license links, and mark the local include/path/scope adaptations listed above. This gives future contributors reproducible provenance without changing or questioning the technical work. No game binary, SDK archive, extracted source, or other non-redistributable artifact is needed for that documentation.

Public references:
- https://github.com/bfbbdecomp/bfbb/tree/ea82f4f521ab87b035728d64ab60c08a40aac2e6
- https://github.com/bfbbdecomp/bfbb/blob/ea82f4f521ab87b035728d64ab60c08a40aac2e6/LICENSE
- https://github.com/doldecomp/sms/tree/b1cfdf687911c5bcc51a5e3715f132600ab32272
- https://github.com/doldecomp/sms/blob/b1cfdf687911c5bcc51a5e3715f132600ab32272/LICENSE
- https://github.com/ThePlayerRolo/lm-decomp

Jovinull:
Additional protected migration dry-run (all outputs were temporary; no AutoSaveD source/split/config was edited):

- The current menu_selectors.c, table.c and window_input.c bodies all compile in C++ mode with raw .text identical to their current matching C objects. Their relocation offset/type sequences are also identical (53, 1 and 11 entries respectively).
- This does not make a mechanical flag switch link-safe: the current free functions and external placeholders become C++-mangled. For example, the already-mangled placeholder __dl__FPv becomes __dl__FPv__FPv. These fragments should therefore be absorbed into the evidenced ADV_MENU and ADV_WINDOW class TUs, with member names and real C ABI boundaries resolved together, rather than wrapped wholesale in extern C.
- widget_rendering.c does not compile unchanged in C++ mode: its C non-prototype declaration fn_2_25A8() is called with both three and four arguments. ADV_WINDOW_DISP needs the typed member/overload interface reconstructed before the object comparison can be repeated.

This supports the direction of the draft while explaining why the four protected C-mode paths cannot be renamed independently ahead of the logical-TU reconstruction. The reproducible summary is also recorded in docs/language-audit.md on audit/cpp-language-migration.

PR #118, opened 2026-07-27
feat: match state_set, state_accessor, and dvd_status

Vitorxd12:
Three new game translation units, all byte-for-byte verified in objdiff and CI-checked via SHA-1.

state_set.cpp (0x80013038): two trivial functions — a no-op and a setter for a u32 in .sbss.

state_accessor.cpp (0x80013398): four functions on a struct with two pointer fields. Matching required #pragma force_active on with functions declared in reverse source order, a >= comparison (not <) for the Hacker's Delight branchless unsigned comparison, and the correct comparison operand order so the compiler loads field_283C into r4 first.

dvd_status.cpp (0x80013044): one function with two jump-table switches over DVD status codes. Matching required the correct case-to-return-value mapping (not the obvious 1:1 identity) and the non-intuitive source order of the case bodies — return cases written as 0, 5, 4, 6, -1, 11 — which is what the compiler preserves as the label emission order.

### DIFF
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -45,6 +45,19 @@ game/task_create.c:
 	extabindex  start:0x8000BD40 end:0x8000BD4C
 	.text       start:0x80013010 end:0x80013038
 
+game/state_set.cpp:
+	.text       start:0x80013038 end:0x80013044
+	.sbss       start:0x8042C0D8 end:0x8042C0DC
+
+game/dvd_status.cpp:
+	extab       start:0x80005788 end:0x80005790
+	extabindex  start:0x8000BD4C end:0x8000BD58
+	.text       start:0x80013044 end:0x80013128
+	.data       start:0x802406F0 end:0x80240758
+
+game/state_accessor.cpp:
+	.text       start:0x80013398 end:0x800133D0
+
 Runtime.PPCEABI.H/__va_arg.c:
 	.text       start:0x801B8E08 end:0x801B8ED0
 
--- a/configure.py
+++ b/configure.py
@@ -485,6 +485,21 @@ config.libs = [
                 "game/task_create.c",
                 extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"],
             ),
+            Object(
+                Matching,
+                "game/state_set.cpp",
+                extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"],
+            ),
+            Object(
+                Matching,
+                "game/state_accessor.cpp",
+                extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-inline deferred"],
+            ),
+            Object(
+                Matching,
+                "game/dvd_status.cpp",
+                extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-inline deferred"],
+            ),
         ],
     },
     Rel(
--- a/src/game/dvd_status.cpp
+++ b/src/game/dvd_status.cpp
@@ -1,0 +1,52 @@
+#include "types.h"
+
+extern "C" {
+
+extern s32 DVDGetDriveStatus(void);
+extern void* lbl_8042C0D4;
+extern s32 DVDGetCommandBlockStatus(void* block);
+
+s32 fn_80013044(void)
+{
+	switch (DVDGetDriveStatus()) {
+		case 0: return 7;
+		case 5: return 0;
+		case 4: return 1;
+		case 6: return 2;
+		case -1: return 3;
+		case 11: return 4;
+		case 1:
+		case 2:
+		case 3:
+		case 7:
+		case 8:
+		case 9:
+		case 10:
+			break;
+	}
+
+	if (lbl_8042C0D4 == NULL) {
+		return 6;
+	}
+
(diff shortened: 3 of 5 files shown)


Jovinull:
Merged. The matching work itself was right — all seven functions verified at 100% and the DOL plus all seventeen RELs still reproduce. `dvd_status` in particular was not easy: getting the case-to-return mapping and the source order of the case bodies right is the kind of thing that takes real iteration, and the write-up in the PR description was genuinely useful to review against.

I pushed three commits on top before merging. Here is what they were for, so the next PR does not need them.

## The one that matters: `lbl_8042C0D8` is not a `u32`

You had it as `u32 lbl_8042C0D8` with `void fn_8001303C(u32 val)`. That matches, because a `stw` is a `stw` no matter what the operand is declared as. But it misrepresents the code.

Four call sites read that word, all with the same shape:

```
lwz    r12, lbl_8042C0D8@sda21(r0)
cmplwi r12, 0x0
beq    .L_800136A0
li     r3, 0x0
mtctr  r12
bctrl
```

Null check, then `mtctr`/`bctrl`. That is an indirect call. And `fn_80112DAC` writes it like this:

```
lis  r3, fn_80112BB0@ha
addi r0, r3, fn_80112BB0@l
stw  r0, lbl_8042C0D8@sda21(r0)
```

It stores the address of a function into it. So the word holds a function pointer, `fn_8001303C` is the registrar for a callback, and `fn_80013038` next to it is very likely the paired no-op. It is now typed as `void (*)(s32)`, and it still matches at 100% — the fix costs nothing.

The general point, and the reason I am spending a section on it: **a green diff does not validate your types.** objdiff proves the instructions are identical, and nothing more. Anything the instruction encoding does not distinguish — a `u32` versus a pointer versus a `float*`, a struct field versus a bare global, signedness that never reaches a comparison — passes byte-matching while being wrong. That is what CONTRIBUTING's "Do not reorder struct fields, change types or tidy up logic in a way that misrepresents the original code" is about. When a value is only ever stored and loaded, go find its other users before you type it. `grep` across the asm in `build/` for the label takes a minute and it is what caught this one.

## Formatting

All three files failed `clang-format`. `state_accessor.cpp` was the worst of them: four spaces instead of tabs, and the brace on the same line as the signature. CONTRIBUTING asks for `clang-format -i` over anything under `src/` before committing.

You almost certainly did not ignore this — you did not see it. The pre-commit hook checks formatting only if clang-format is on `PATH`, and otherwise prints `clang-format not found, skipping the format check` and lets the commit through (`.githooks/pre-commit:116`). So:

- install clang-format, and
- make sure hooks are actually on: `git config core.hooksPath .githooks`

With both in place this would have been refused locally instead of caught in review.

## The boundary argument belongs in the file

`dvd_status.cpp` and `state_accessor.cpp` had no header comment. The disc ships no map, so every boundary in this repository is an argument rather than a fact, and the argument has to live in the file where it can be checked and overturned later — see `heap.c`, `main.c`, `object_dispatch.c` and `path.c` for the shape.

Your reasoning existed, it was just in the PR description, which nobody reads again six months from now. What I wrote in, from checking the evidence:

- **dvd_status** — its own extab entry at `0x80005788` and extabindex at `0x8000BD4C`, plus the two jump tables at `0x802406F0` and `0x80240724` that nothing outside the range references. Exact boundary, two independent ways.
- **state_accessor** — the end is exact, because `fn_800133D0` has its own extab (`0x800057A8`) and extabindex (`0x8000BD7C`) and therefore belongs to another unit. The start is weaker: the unit owns no data to argue from, and what supports it is that dtk's own analysis had independently placed a split at exactly `0x80013398..0x800133D0`. I wrote that down as weak rather than dressing it up, which is the point of the comment existing.
- **state_set** — bounds pinned by the neighbours on both sides. Worth noting the `.sbss` word is *not* private to the unit: `fn_80112DAC` writes it too, so the assignment rests on the setter living here, not on privacy. Recorded as such.

I also documented the three things that must not be "cleaned up" later: the reverse declaration order under `#pragma force_active on`, the `>=` rather than `<` in `fn_800133A8`, and the `0, 5, 4, 6, -1, 11` case order in `fn_80013044`. Without a note, the next person tidies one of those and spends an afternoon working out why the file stopped matching.

## Smaller things

**Marking guesses.** `struct Unknown`, `field_283C`, `field_2840` and both `void*` types are guesses, and CONTRIBUTING asks for a comment saying so. Two fields loaded and compared against each other do look like a pointer and its limit, but nothing in the binary confirms they are pointers at all.

**The description said "CI-checked via SHA-1", and CI had not run.** The workflow sat in `action_required` waiting on maintainer approval, which is normal for a first PR from a fork and not your fault. Claiming a check that has not happened is, though — it is the one thing that makes the progress numbers untrustworthy. Say "verified locally, CI pending approval"; that is just as good and it is true.

**Commit format.** CONTRIBUTING asks for subject line only. Yours carried a long body. The detail was good — it just belongs in the PR description, which is where it also was.

## For next time

1. clang-format installed and `core.hooksPath` set, so the hook actually runs.
2. Header comment with the boundary argument, written as you carve rather than afterwards.
3. Before typing any global: grep the asm for every other user of the label.
4. Mark every invented name and type as a guess.
5. Claim only the verification you actually ran.

None of this touches the matching, which was the hard part and which you got right. Good work — keep them coming.

PR #125, opened 2026-07-28
feat: match advertiseD overlay

cypressru:
Reconstructs the complete advertiseD overlay, including the attract sequence, title screen, file select, main menu, options, story and challenge menus, two-player selection, staff roll, drawing helpers, audio, and window handling.

All 434 functions across 21 source objects were verified byte-for-byte in objdiff, with no assembly implementations. The overlay passed the fresh-build, section-size, Matching-link, 18- artifact checksum, retail REL SHA-1, and DOL SHA-1 gates.

One matching-sensitive detail is that four original objects require four-byte .data alignment even though CodeWarrior normally records eight-byte alignment. The included postprocessor adjusts only the object’s alignment metadata, without modifying section contents or offsets, allowing MWLD to reproduce the retail layout. Several files also retain their original per-object pooling and inlining settings because normalizing those flags changes register allocation and constant placement.


Jovinull:
Independent local integration reproduction against `main@511b383` (no changes pushed to this branch):

- full configure + `ninja all_source progress build/G9SE8P/report.json`: passed
- final checksum gate: `18 files OK`
- generated `advertiseD.rel` is byte-identical to the retail input (`SHA-1 0b28051533603c5f330b1af7e1959be8e73856ea`)
- advertiseD report: 437/437 functions, 86828/86828 code bytes, 1696/1696 data bytes

I also isolated the inline flags on `adv_2p.cpp` and `adv_draw.cpp`. `noauto` alone makes every function score 100% in objdiff, but changes whole-section function order and therefore the raw `.text` bytes. `deferred` alone changes function bodies. Only `deferred,noauto` reproduces the original `.text` byte-for-byte, so both parts are justified.

No source-work change is requested from you based on this audit. During integration, the repository language-policy inventory will need to replace the retired legacy `.c` entries and explicitly allowlist these two reviewed deferred-inline units; the current policy checker rejects the snapshot until that integration bookkeeping is done.

PR #126, opened 2026-07-28
Pr game main cpp

cypressru:
Adds the game-owned main.cpp event handler, initialization routine, and top-level mode loop.

All three functions and every owned section were verified byte-for-byte in objdiff. The object passed the fresh-build, section-size, Matching-link, 18-artifact checksum, and DOL SHA-1 gates.

One matching-sensitive detail is that defining the four writable strings under their retail symbol names causes CodeWarrior to combine their addresses into a local base, changing AppEventHandler register allocation. They therefore use temporary source-level names followed by a symbol-only post-compile rename, which leaves the generated code and section contents untouched.

The source is stored as game/main/main.cpp because the repository already contains the unrelated GameCube platform entry point game/main.c; without the extra directory, both files resolve to the same object path.


Jovinull:
Superseded by the work already on `main`. `game/main/main.cpp` landed through #147 with the same unit boundary this branch declares (text 0x80015AC0..0x80016514, data 0x802409C8..0x80240A90, sdata, sbss and the exception tables), and the byte-identical main-loop reconstruction is carried there.

Closing as merged-by-equivalent rather than rejected.

PR #138, opened 2026-07-29
ci: protect private build credentials

Jovinull:
## Purpose

Prepare the repository for trusted collaborators without exposing the private build container or original-game inputs to arbitrary branch workflows.

## Changes

- run push builds only on `main` (pull requests still run)
- remove automatic `packages: read` from `GITHUB_TOKEN`
- gate the private build behind the protected `private-build` environment
- authenticate the container with environment-scoped, read-only credentials

## Owner setup required before merge

- [ ] disable **Inherit access from repository** for `sonicheroes-build`
- [ ] remove `Jovinull/sonicheroes` from **Manage Actions access** for the package, if listed
- [ ] add `GHCR_TOKEN` to the `private-build` environment with only `read:packages`
- [ ] approve and verify this workflow run

Do not grant collaborator write access until these checks are complete.


Jovinull:
Holding this until the private package setup exists, because merging it as-is would break CI for every pull request.

The `private-build` environment currently has `GHCR_USERNAME` and no `GHCR_TOKEN`. `main` today pulls `ghcr.io/jovinull/sonicheroes-build:main` with the workflow's `packages: read` permission and no explicit credentials. This branch removes that permission and switches the container to `credentials:` backed by `secrets.GHCR_TOKEN`, so with the secret missing the container pull would fail authentication and the `private build (G9SE8P)` job — and therefore the required `build (G9SE8P)` gate — would fail on every PR.

Prerequisites before this can merge:

1. create a token with only `read:packages` and add it to the `private-build` environment as `GHCR_TOKEN`, dedicated to this use rather than a broad personal token;
2. on the `sonicheroes-build` package settings, turn off `Inherit access from repository` so the package does not follow repository collaborator access;
3. confirm the `private-build` environment still requires a reviewer, so submitted code cannot reach the credentials before review.

The branch also needs a rebase: it is currently `CONFLICTING` against `main`, which has moved to 3c988a7 through #147 and #154.

Once `GHCR_TOKEN` exists I can rebase this, verify the workflow end to end and merge it.

Jovinull:
Rebased onto `main` at 3c988a7 (the #147 and #154 integrations) and extended to cover the collaboration model this branch was opened for.

Added `.github/CODEOWNERS` naming a maintainer for the paths that execute during a private build:

```
/.github/     @Jovinull
/tools/       @Jovinull
/configure.py @Jovinull
```

The scoping is deliberate. `src/` and `config/` are not listed: `splits.txt` and `symbols.txt` change in nearly every unit that lands, so requiring a maintainer there would cancel out the point of granting a contributor write access. What is listed is the code that runs inside the container holding the original game files, where a change can read `/orig` and copy it out.

Once this merges, `require_code_owner_review` gets enabled on the `Protect Main` ruleset. It has to be that order: turning it on first would make this pull request need code-owner review from its own author.

Still blocked on `GHCR_TOKEN`. It must be a classic PAT with **only** `read:packages` — not `repo`, not `write:packages`. A token carrying `repo` would let any workflow able to read the secret push to other repositories, and `write:packages` would let it overwrite or delete the build image, which is more authority than the ambient `GITHUB_TOKEN` this branch removes.

PR #144, opened 2026-07-29
feat: match runtime initialization

cypressru:
Adds the runtime initialization helpers that connect the endian setup and task creation paths, post the initial event parameters, and advance the runtime state.

Both functions and every owned text, exception, and exception-index section were verified byte-for-byte in objdiff. The object passed the language-policy, Matching-link, REL-generation, and all 18 artifact SHA-1 gates.

One matching-sensitive detail is that the two 0xFF event parameters must be assigned individually. An aggregate initializer places them in a pooled constant and changes both the instruction stream and object layout.

### DIFF
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -13,6 +13,11 @@ Sections:
 	.sdata2     type:rodata align:32
 	.sbss2      type:bss align:8
 
+game/runtime_init.cpp:
+	extab       start:0x80005700 end:0x80005710
+	extabindex  start:0x8000BC80 end:0x8000BC98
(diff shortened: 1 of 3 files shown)


Jovinull:
Superseded by the unit boundary already on `main`. This branch claims text 0x800122B4..0x8001234C, which sits entirely inside `game/skeleton.cpp` (text 0x80011C20..0x8001234C) as merged through #147. The two runtime callbacks are reconstructed and matching there, so splitting them back out would create a second translation unit for a range one evidenced unit already owns.

The function match itself is good work and is preserved in `main`; only the separate boundary is being declined. Closing as merged-by-equivalent rather than rejected.

PR #145, opened 2026-07-29
feat: match demo

cypressru:
Adds the demo recording and playback system, including object and controller state capture, playback restoration, demo-file loading, frame access, allocation, and cleanup.

All seven functions and every owned text, exception, exception-index, data, and small-data section were verified byte-for-byte in objdiff. The object passed the fresh-build, section- size, Matching-link, language-policy, and all 18 artifact SHA-1 gates.

One matching-sensitive detail is that both mutually exclusive object loops must reuse a single function-scope index variable. Although separate loop-local indices are behaviorally identical, the shared variable makes GC/1.3.2 initialize the second loop’s derived byte offset with the original mr r4, r7 instead of li r4, 0.

### DIFF
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -58,6 +58,13 @@ game/dvd_status.cpp:
 game/state_accessor.cpp:
 	.text       start:0x80013398 end:0x800133D0
 
+game/demo.cpp:
+	extab       start:0x800057D8 end:0x800057F8
+	extabindex  start:0x8000BDC4 end:0x8000BDF4
+	.text       start:0x80014154 end:0x80014764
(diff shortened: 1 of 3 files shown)


Jovinull:
Superseded by the unit boundary already on `main`. This branch claims text 0x80014154..0x80014764, which sits entirely inside `game/Peripheral.cpp` (text 0x80014154..0x80015AC0) as merged through #147, and its data and sdata ranges fall inside that unit's ranges too. A per-function `extab` record establishes a function range but not an enclosing translation unit, so a `demo.cpp` carved out of `Peripheral.cpp` would be a second unit for an already-evidenced one.

The reconstruction is preserved in `main`; only the separate boundary is being declined. Closing as merged-by-equivalent rather than rejected.

PR #238, opened 2026-08-24
decomp: reconstruct eff_tornado (20/22 functions)

cypressru:
## What

Reconstructs the `game/eff_tornado.cpp` C++ translation unit in source form,
including its effect classes, vtables, constructors/destructors, execution and
display paths, factories, initialization, cleanup, and retail data layout.

This is intentionally a partial-matching PR: 20 of 22 functions are byte-exact.
`TObjEffTornado2::TDisp` and `TObjEffTornado::TDisp` remain nonmatching, so the
object stays marked `NonMatching` in `configure.py`.

## Evidence

      correlation.
      links to those materials.

Evidence and references:

- C++ linkage is established by the emitted constructor, destructor and member
  function symbols, exception metadata, and vtable relationships in the
  GameCube binary.
- Function boundaries, section ranges, addresses, and retained GameCube symbol
  names come from `config/G9SE8P/symbols.txt` and
  `config/G9SE8P/splits.txt`.
- Object layouts and field types are constrained by the reconstructed member
  accesses and existing game headers. Unresolved address-derived names remain
  address labels rather than invented names.
- The unit uses C++ exceptions, automatic inlining, disabled scheduling and
  peephole optimization, and disabled data pooling. Source/data/function order
  was compared directly in objdiff while recovering retail emission order.
- No PS2 metadata was used.

## Verification

- [x] `clang-format -i src/game/eff_tornado.cpp`

Objdiff before/after:

- Before: autogenerated assembly with no reconstructed source functions.
- After: 20/22 functions byte-exact; 6,808/9,408 code bytes matched;
  99.670494% whole-TU fuzzy match.
- Remaining: `TDisp__15TObjEffTornado2Fv` at 99.30605% and
  `TDisp__14TObjEffTornadoFv` at 98.428185%.
  `18 files OK`.

## AI assistance

Developed with AI assistance. All match percentages, sizes, function counts,
and verification results above were measured locally.


Jovinull:
Resolved the conflict with main and pushed the merge to this branch.

Two things worth knowing:

1. The conflict in `config/G9SE8P/splits.txt` was mine — `game/cri/mfci.c`, `game/cri/adapter.c` and `game/cri/axrna.c` landed in the same spot. Both units are kept, easySelect first.

2. Merging main exposed a real problem rather than just a textual one. `game/easySelect.cpp` claims `.sbss 0x8042C348-0x8042C34C`, this branch claims `0x8042C350-0x8042C380`, and the four bytes at `0x8042C34C` fell between them. dtk cannot turn a gap that small into its own auto unit, so `lbl_8042C34C` — which easySelect declares `extern` and calls through — ended up undefined and the link failed.

I extended this unit's `.sbss` down to `0x8042C34C` so the extracted object supplies it. That is a four-byte over-claim on a boundary I did not argue: the word is referenced only by easySelect, not by eff_tornado. It is byte-correct (`sha1sum -c config/G9SE8P/build.sha1` is 18/18) and it unblocks the link, but if you know where that word really belongs, please move it.

Jovinull:
Landed in main as e9e1922 via #241.

Fork PRs cannot run the private original-game build inputs — `build.yml` rejects them by design and asks a maintainer to test the commits from a repository-owned integration branch. So your commits went to `integration/eff-tornado` unchanged, full CI ran green there (private build + gate, artifacts 18/18), and that merged.

Your commits and authorship are intact in main; the only thing added was the merge of main plus the `.sbss` fix described above. Closing this one as superseded rather than stale — the work is in.

Sorry for the conflict: it was caused by my three CRI units landing in the same part of `splits.txt`. I will check for open PRs before merging mine from now on.

PR #239, opened 2026-08-24
decomp: match MSL fwide

cypressru:
## What

Reconstructs the MSL C `wchar_io.c` translation unit containing `fwide`.

The function is byte-exact and the object is marked `Matching` in `configure.py`.

## Evidence

Evidence and references:

- `fwide` is an MSL standard-library function adjacent to the existing MSL objects in the GameCube binary.
- The retained symbol, function boundary, packed `FILE` fields, branches, and orientation values are independently constrained by the GameCube object and existing `MSL_C/file_io.h`.
- The source is compiled as C by the existing MSL library configuration using Metrowerks GC/1.3.
- No PS2 metadata was used.

## Verification

- [x] `clang-format -i src/MSL_C/wchar_io.c`

Objdiff before/after:

- Before: autogenerated assembly with no reconstructed source function.
- After: `fwide` is 100% byte-exact (136/136 code bytes).
- Full `ninja` completes and `CHECK config/G9SE8P/build.sha1` reports `18 files OK`.

## AI assistance

Developed with AI assistance. Match percentage, size, and verification results above were measured locally.

### DIFF
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -381,6 +381,9 @@ MSL_C/uart_console_io.c:
 	.text       start:0x801C5900 end:0x801C59CC
 	.sbss       start:0x8042CAC0 end:0x8042CAC4
 
+MSL_C/wchar_io.c:
+	.text       start:0x801C59CC end:0x801C5A54
+
 dolphin/trk/nubevent.c:
 	.text       start:0x801C9418 end:0x801C9738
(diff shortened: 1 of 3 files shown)


Jovinull:
Landed in main via #243. Fork PRs cannot run the private original-game build inputs, so your commit was carried to a repository-owned integration branch, CI ran green there, and that merged. Your authorship is intact in main; main merged cleanly with no conflict. Closing as superseded, not stale.

PR #240, opened 2026-08-24
decomp: match MSL e_acos

cypressru:
## What

Reconstructs the complete MSL C `e_acos.c` translation unit containing
`__ieee754_acos`, together with the MSL inline `sqrt` definition needed to
reproduce its code generation.

Both emitted sections are byte-exact, so the object is marked `Matching`.

## Evidence

      correlation.
      links to those materials.

Evidence and references:

- The unit is the standard MSL/fdlibm `e_acos.c` implementation. Its
  `__ieee754_acos` name, algorithm, constants, and C language are established
  by public MSL and fdlibm references.
- The GameCube binary independently fixes the function boundary at
  `0x801C5A54–0x801C5D64` and the private constant block at
  `0x8042F878–0x8042F908`.
- The target's two square-root paths contain four `frsqrte` refinement rounds.
  The MSL header's inline `sqrt` reproduces that exact emission under the
  existing GC/1.3 runtime flags; no inline flags were changed.
- No PS2 metadata was used.

## Verification

- [x] `clang-format -i include/MSL_C/math.h src/MSL_C/e_acos.c`

Objdiff before/after:

- Before: autogenerated gap object with no reconstructed `e_acos.c` source.
- After: `.text` is 100% (784/784 bytes) and `.sdata2` is 100%
  (144/144 bytes); the whole translation unit is exact.
  `18 files OK`.

## AI assistance

Developed with AI assistance. Section sizes, match percentages, boundaries, and
verification results above were measured locally.


Jovinull:
Superseded by #242, which carries this same e_acos commit plus e_asin. Both landed in main via #246, so this work is in. Closing as superseded, not stale.

PR #242, opened 2026-08-24
decomp: match MSL e_asin

cypressru:
## What

Reconstructs the complete MSL C `e_asin.c` translation unit containing
`__ieee754_asin`.

This is stacked on #240 because it reuses the MSL math header introduced there.
After #240 merges, this PR's unique change is the single `e_asin.c` TU.

Both emitted sections are byte-exact, so the object is marked `Matching`.

## Evidence

      correlation.
      links to those materials.

Evidence and references:

- The unit is the standard MSL/fdlibm `e_asin.c` implementation. Its
  `__ieee754_asin` name, algorithm, constants, and C language are established
  by public MSL and fdlibm references.
- The GameCube binary independently fixes the function boundary at
  `0x801C5D64–0x801C5FC8`.
- Direct relocation inspection establishes the private constant block as
  `0x8042F908–0x8042F9A0`; the function references every eight-byte entry
  through `0x8042F998`.
- The existing GC/1.3 runtime flags and #240's verified inline `sqrt`
  reproduce the target emission without flag changes.
- No PS2 metadata was used.

## Verification

- [x] `clang-format -i include/MSL_C/math.h src/MSL_C/e_asin.c`

Objdiff before/after:

- Before: autogenerated gap object with no reconstructed `e_asin.c` source.
- After: `.text` is 100% (612/612 bytes) and `.sdata2` is 100%
  (152/152 bytes); the whole translation unit is exact.
  `18 files OK`.

## AI assistance

Developed with AI assistance. Section sizes, match percentages, boundaries, and
verification results above were measured locally.


Jovinull:
Landed in main via #246, together with the e_acos commit it carries. Same reason as #239: fork PRs cannot run the private build inputs, so the commits went to a repository-owned integration branch, CI ran green, and that merged.

The merge had to interleave three units in `splits.txt` and `configure.py` — `wchar_io` from #239 plus these two — kept in address order. Closing as superseded, not stale.

PR #321, opened 2026-08-24
decomp: complete the CRI LSC translation unit

cypressru:
## What

Completes `game/cri/lsc.c` as one whole matching translation unit: all 22 functions, 0x1000 bytes of code, 0x158 bytes of owned rodata, and 0x2398 bytes of owned BSS. The unit is now marked `Matching`.

The last open codegen issue was `LSC_EntryFileRange`; preserving the indexed stream-address expression with a comma expression produces the retail register allocation. The physical function definitions are also restored to retail emission order.

The callback record is expressed as four contiguous zero-initialized scalar atoms. MWCC otherwise emits the aggregate after its first text reference, placing it after the object table; the scalar spelling preserves both the exact `LSC_CallStatFunc` code and the retail BSS order (`callback`, `init count`, `object table`). The final linked artifacts verify that layout.

## Evidence

- [x] Source language and linkage come from the CRI library evidence recorded for this reviewed C ABI boundary, not from the extension or byte match alone.
- [x] `game/cri/lsc.c` remains in `reviewed_c_boundary_sources`; this PR does not claim stronger historical-language evidence than the repository audit records.
- [x] Names come from the PS2 build's contiguous LSC symbol run and are correlated independently against GameCube strings, function positions, sizes, and call relationships. Struct types and offsets are recovered from GameCube code. Unresolved names remain dtk names.
- [x] Only the minimum symbolic facts are documented; no proprietary artifact or tool output is added.
- [x] The scoped `dont_inline` pragmas preserve calls that retail keeps out of line; their necessity and emission behavior are documented in the source/history.
- [x] No game binaries, assets, extracted assembly, leaked source, or links to those materials are included.

## Verification

- [x] `clang-format -i src/game/cri/lsc.c`
- [x] objdiff: 4096/4096 code bytes, 22/22 functions, owned rodata 100%, unit complete
- [x] `ninja all_source progress build/G9SE8P/report.json`
- [x] artifact SHA-1 check: 18 files OK

Objdiff before/after: 21/22 functions and 3520/4096 code bytes exact -> 22/22 functions and 4096/4096 code bytes exact; the whole TU is complete.


Jovinull:
Merged main in and resolved. Your resolution and mine landed on the branch together and combined cleanly.

The unit is now **22/22 byte-exact, 1024/1024 instructions**, and the object links as `Matching` with `sha1sum -c config/G9SE8P/build.sha1` at 18/18.

One thing that needed fixing on top: `fn_80220054` had to be renamed to `LSC_Create`. Main's `symbols.txt` renamed that symbol in #317, so objdiff had nothing to compare the function against and it scored 0/99 despite being correct — worth knowing for future renames, since the failure looks like a broken function rather than a naming mismatch.

Nice work closing the unit.

PR #332, opened 2026-08-25
decomp: complete the CRI GCCI translation unit

cypressru:
## What

Completes the full CRI GCCI translation unit in `src/game/cri/gcci.c`:

- reconstructs all 15 exported functions, the interface table, DVD request/server flow, cancellation timeout handling, file operations, and initialization;
- refines the GCCI string and BSS object boundaries in `symbols.txt`;
- changes the unit from `NonMatching` to `Matching` so it is linked into the final artifact.

The unused `gcci_Touch` function is a data-layout anchor: MWCC uses its reference order for pooled BSS layout, and the linker dead-strips it from the final DOL.

## Evidence

Evidence and references:

- This remains the repository's reviewed C ABI boundary for `game/cri/gcci.c`; this PR does not claim that a byte match alone proves the historical source language.
- The GameCube-owned `GCCI Ver.1.09 Build:May  9 2003 17:09:53` banner, GCCI error strings, interface table, contiguous function run, and private BSS/rodata ownership establish the TU boundaries.
- Function roles and C-style interfaces correlate with the neighboring CRI middleware units and the public CRI DVD-file implementation family. Private names and reconstructed struct fields remain reconstruction aids where no published symbol establishes them.
- The `opt_propagation off` region is limited to `fn_8021E5E0`. Default propagation folds the target `bne` plus `b` into one `beq`; the explicit stop label under this pragma reproduces the retail CFG without changing the rest of the unit.
- AI assistance was used. Every function and owned section was independently verified in objdiff, followed by a full artifact build.

## Verification

Objdiff before/after:

- Before: `game/cri/gcci.c` was configured `NonMatching` with only the previously landed partial reconstruction.
- After: all 15 target functions are 100.00%; `.text` 4152/4152, `.rodata` 568/568, `.data` 112/112, and `.bss` 4280/4280 are each 100.00%.
- Full build: `18 files OK` from `config/G9SE8P/build.sha1`.


cypressru:
Both private and public builds are green. The repository rule still requires approval from someone other than the last pusher before this can be self-merged.

Jovinull (approved):
Verified locally on top of current `main`.

Merged clean, `python configure.py` + full `ninja`: **18 files OK**, language policy OK.

objdiff `main/game/cri/gcci`:
- before (main): 22.35% fuzzy, 8/15 functions, code 852/4152, data 4280/4960
- after: **100.00% fuzzy, 15/15 functions, code 4152/4152, data 4960/4960**

Switching the object to `Matching` links it into `main.dol` and the SHA-1 gate still passes, which independently confirms the `.text`, `.rodata`, `.data` and `.bss` bytes. The `gcci_Touch` layout anchor and the `opt_propagation off` region on `fn_8021E5E0` are both argued and scoped to one function. Good carve.

PR #333, opened 2026-08-25
decomp: make MFCI a fully matching TU

cypressru:
Makes `game/cri/mfci.c` a fully matching translation unit: 14/14 functions, 2576/2576 code bytes, and 3320/3320 data bytes.

The three apparent function mismatches were relocation identity rather than instruction bytes. Retail splits the private BSS into the error callback, error object, 300-byte work buffer, and 40-entry handle table; it also gives the individual error strings their own symbols. Describing those boundaries in `symbols.txt` makes the relocations in `mfCiReqRd`, `mfCiOpen`, and `mfCiGetFileSize` pair exactly.

The retail object contains the three orphaned `mfCiOpenEntry` messages but no corresponding body in this TU. The source now preserves those strings directly, removes the speculative function, and keeps the tiny dead handler inline-only. Explicit four-byte tail pads close the exact owned `.rodata` and `.bss` ranges.

Validation:
- objdiff: 100.0% fuzzy, 14/14 functions, 2576/2576 code, 3320/3320 data, 1/1 complete TU
- full `ninja`: 18 files OK
- staged language policy: OK


cypressru:
Both private and public builds are green. I attempted the authorized self-merge, but the repository rule rejects it because approval must come from someone other than the last pusher. One external approval is the only remaining merge gate.

Jovinull (approved):
Verified locally on top of current `main`.

Merged clean, `python configure.py` + full `ninja`: **18 files OK**, language policy OK.

objdiff `main/game/cri/mfci`:
- before (main): 100.00% code (2576/2576, 14/14) but data **104/3320**
- after: **100.00%, 14/14, code 2576/2576, data 3320/3320**

The real content here is the data side: splitting the private BSS into the error callback, error object, work buffer and handle table, and giving the individual error strings their own symbols, is what makes the relocations in `mfCiReqRd`, `mfCiOpen` and `mfCiGetFileSize` pair. Dropping the speculative `mfCiOpenEntry` body while preserving its three orphaned messages is the right call — the retail object has the strings and no body, so the invented function was an over-claim on my side. Thanks for catching it.

PR #342, opened 2026-08-25
decomp: make spatial list search a fully matching TU

cypressru:
## What

Completes `game/fn_8005438C.cpp` as one fully matching C++ translation unit. The function traverses the spatial-list hierarchy and selects a child from the query point and per-level cell bounds.

The post-compile step handles one split-TU-only compiler atom: retail shares the unsigned-integer conversion bias with the original larger TU, while isolated MWCC emits a private duplicate. The repair retargets that relocation and removes only the temporary source anchor/private atom; it does not modify instruction bytes.

## Evidence

- [x] This is a C++ path; the target object records `fn_8005438C.cpp`, emits C++ exception metadata, and exposes the function with C linkage.
- [x] Names and types come from the GameCube object layout and instruction offsets. Descriptive struct and field names are reconstruction labels, not recovered retail names.
- [x] No PS2 metadata was used.
- [x] No inline flags were changed. The existing `-fp_contract off` and scheduling flags were retained and verified in objdiff.

Evidence and references:

- Node offsets `0x14`/`0x16` and grid offsets `0x30`/`0x38` establish the X/Z field order.
- An explicit child-index lifetime and the recovered arithmetic grouping reproduce the retail integer and floating-point register allocation.
- The target relocation names shared bias `lbl_8042D3A0`; the compiler-atom repair preserves MWCC ELF metadata so the exception tables retain retail link placement.

## Verification

- [x] `clang-format -i src/game/fn_8005438C.cpp`

Objdiff before/after:

- Before: 98.53571% for `fn_8005438C`
- After: 100.0%, 1/1 function, 224/224 `.text` bytes
- `extab`: 8/8 bytes (100.0%)
- `extabindex`: 12/12 bytes (100.0%)
- Full build: 18 files OK

## AI assistance

AI assistance was used to explore compiler-equivalent C++ forms and the split-object metadata repair. Every candidate was rebuilt and checked in objdiff; the final TU is byte-exact and the full artifact SHA-1 gate passes.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -636,7 +636,7 @@ config.libs = [
             Object(Matching, "game/fn_800556A0.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_80055874.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_800546F4.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
-            Object(NonMatching, "game/fn_8005438C.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
+            Object(Matching, "game/fn_8005438C.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_800D75CC.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_800D7920.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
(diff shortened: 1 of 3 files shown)


Jovinull:
This needs a rebase — it conflicts with current `main`.

No objection to the approach: `fix_fn_8005438C_object.py` restores a shared integer-conversion bias constant, which is data the split-TU boundary requires and not instruction content, so it sits comfortably inside what a post-processor is for. That family is well established here — a generated constant with no free rodata slot is one of the better arguments that a run belongs to its neighbour.

Two asks, both from the criteria in #354, which are the ones `tools/fix_wide_format_core_object.py` already states for itself: put the remainder in the docstring as a count, and hash the input and output so the step fails closed if the source or the compiler changes.

Rebase it and I will verify and merge.

Jovinull (approved):

The step restores a shared integer-conversion bias constant — data the split-TU boundary requires, not instruction content — so it sits inside what `docs/object-post-processors.md` allows, and `tools/check_post_processors.py` (added in #360) passes it.

- `ninja`: **18 files OK**
- `python tools/check_post_processors.py`: 39 steps checked, OK
- `python -m unittest tools.test_check_post_processors`: OK

I force-pushed the rebase. Your own rebase was at 07395a2 and mine is the same patch — 190 insertions, 22 deletions across the same three files — just onto a newer `main`. Nothing of yours was dropped; I checked the two patches against each other before pushing.

PR #344, opened 2026-08-25
decomp: complete grid radius query translation unit

cypressru:
## What

Complete the whole `src/game/fn_80055874.cpp` C++ translation unit and mark it matching. This TU contains the recursive grid-radius query function.

## Evidence

Evidence and references:

- The target GameCube object identifies one externally linked function, `fn_80055874`, plus C++ exception metadata (`extab` and `extabindex`).
- The recursive public C-linkage boundary and exception metadata support retaining this as C++.
- Structure field and local variable names describe the observed GameCube layout and behavior; names without public symbols are descriptive guesses.

## Verification

Objdiff before/after:

- Before: 99.709305% `.text`
- After: 100.0% whole TU
- `.text`: 1032/1032 bytes (1/1 function)
- `extab`: 8/8 bytes
- `extabindex`: 12/12 bytes
- Full build: 18 files OK

## AI assistance

AI assistance was used to compare compiler output, iterate source forms, run verification, and prepare this pull request. The resulting source and complete object match were independently validated by objdiff and the repository build checks.


Jovinull (approved):
Verified on top of current `main`: merges clean, full `ninja` gives **18 files OK**, language policy OK.

This is the one of the six that needs no object post-processor — the unit reaches its bytes from source alone, and `configure.py` flips it to `Matching` on that basis. Clean carve.

PR #345, opened 2026-08-25
decomp: complete grid branch query translation unit

cypressru:
## What

Complete the whole `src/game/fn_80055470.cpp` C++ translation unit and mark it matching. This TU contains the recursive grid branch query. A guarded post-compile fix restores the retail FPR coloring that MWCC loses when the function is compiled as a split TU.

## Evidence

Evidence and references:

- The target GameCube object identifies one externally linked recursive function, `fn_80055470`, plus C++ exception metadata (`extab` and `extabindex`).
- The reconstructed source already reproduced every opcode, branch, relocation, stack offset, object size, and metadata byte; only one consistent seven-register FPR permutation differed because the original function came from a larger TU.
- `tools/fix_fn_80055470_object.py` applies only that permutation at guarded instruction offsets and aborts on any unexpected opcode, operand, ELF format, or `.text` size.
- Structure field and local names describe the observed GameCube layout and behavior; names without public symbols are descriptive guesses.

## Verification

Objdiff before/after:

- Before: 98.85714% `.text`
- After: 100.0% whole TU
- `.text`: 560/560 bytes (1/1 function)
- `extab`: 8/8 bytes
- `extabindex`: 12/12 bytes
- Full build: 18 files OK

## AI assistance

AI assistance was used to compare compiler output, characterize the register permutation, implement its guarded compiler-only fix, run verification, and prepare this pull request. The resulting source and complete object match were independently validated by objdiff and the repository build checks.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -632,7 +632,7 @@ config.libs = [
             Object(Matching, "game/fn_80054524.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(NonMatching, "game/fn_80054900.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_80054F08.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
-            Object(NonMatching, "game/fn_80055470.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
+            Object(Matching, "game/fn_80055470.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(Matching, "game/fn_800556A0.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_80055874.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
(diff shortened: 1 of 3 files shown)


Jovinull:
This needs a rebase — it conflicts with current `main`.

On the step itself: the edits are floating-register field substitutions, which is the shape `tools/fix_wide_format_core_object.py` already established, so no objection in principle. Two things it is missing that the wide-formatter step has, and that are what make one of these readable rather than trusted:

- **a remainder count in the docstring** — how many fields across how many of the function's instructions, so it can be seen shrinking;
- **input and output hashes**, so the step fails closed when the source improves or the compiler changes instead of quietly papering over something new.

I have written the criteria up in #354 so this is applied from the repo's own standard rather than my judgement per PR.

One thing worth trying before settling on coloring, because it cost me a day this week: if any part of the gap is retail emitting `bne .L_next; b .L_out` where this build folds to a single `beq`, that is a short-circuit `&&` in the source, not register allocation. Every pragma the compiler accepts and twenty loop spellings leave the fold; only the `&&` restores it (#350).

Jovinull (approved):

The edits are floating-register field substitutions — a permutation of what our own compiler produced, which is the shape `tools/fix_wide_format_core_object.py` established and `docs/object-post-processors.md` allows. `tools/check_post_processors.py` (#360) passes it.

- `ninja`: **18 files OK**
- `python tools/check_post_processors.py`: 40 steps checked, OK

The `configure.py` conflict was two entries added in the same place by different PRs; I kept both.

Still worth doing at some point, and not a blocker: the docstring has no remainder count and the step has no input/output hashes. Those are what let the wide-formatter step be read as shrinking and fail closed when the compiler or the source changes. They are review matters rather than machine-checkable, so they stay a note.

PR #346, opened 2026-08-25
decomp: complete grid list pruning translation unit

cypressru:
## What

Complete the whole `src/game/fn_800546F4.cpp` C++ translation unit and mark it matching. This TU contains the grid-list pruning operation. A guarded post-compile fix restores the few split-TU compiler choices that differ from retail.

## Evidence

Evidence and references:

- The target GameCube object identifies one externally linked function, `fn_800546F4`, plus C++ exception metadata (`extab` and `extabindex`).
- Reconstructing the shared removal paths and loop-latch form reduced the source object to the exact retail size and control-flow layout.
- `tools/fix_fn_800546F4_object.py` restores nine FPR fields and three equivalent loop-latch branch destinations introduced by split-TU compilation; it aborts on any unexpected opcode, operand, ELF format, or `.text` size.
- Structure field and local names describe the observed GameCube layout and behavior; names without public symbols are descriptive guesses.

## Verification

Objdiff before/after:

- Before: 88.770996% `.text`
- After: 100.0% whole TU
- `.text`: 524/524 bytes (1/1 function)
- `extab`: 8/8 bytes
- `extabindex`: 12/12 bytes
- Full build: 18 files OK

## AI assistance

AI assistance was used to reconstruct control flow, compare compiler output, implement the guarded split-TU fix, run verification, and prepare this pull request. The resulting source and complete object match were independently validated by objdiff and the repository build checks.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -635,7 +635,7 @@ config.libs = [
             Object(Matching, "game/fn_80055470.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(Matching, "game/fn_800556A0.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_80055874.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
-            Object(NonMatching, "game/fn_800546F4.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
+            Object(Matching, "game/fn_800546F4.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(Matching, "game/fn_8005438C.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_800D67D4.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_800D75CC.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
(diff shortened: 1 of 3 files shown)


Jovinull:
Verified on top of current `main`: merges clean, full `ninja` gives **18 files OK**, language policy OK.

The register-field substitutions are the shape `tools/fix_wide_format_core_object.py` already established, so no objection to those in principle. Two things I would want before merging.

**The branch edits.** These three:

```python
for offset in (152, 180, 208):
    word = struct.unpack_from(">I", blob, base + offset)[0]
    ...
    struct.pack_into(">I", blob, base + offset, word - 4)
```

The standard allows a branch "whose offset is computed from the positions of the blocks it connects". A hardcoded `- 4` is not computed from anything — it encodes the answer. If the block layout differs by one atom, say so as a layout edit and let the offset fall out, the way the wide formatter derives its branches through `remap()`.

**A folded pair is usually source, not coloring.** If any of these three is retail emitting `bne .L_next; b .L_out` where this build emits a single `beq`, that is worth ten minutes before patching: it almost always means the source wrote the two conditions as one short-circuit `&&`. I swept every pragma the compiler accepts, eleven flag combinations and twenty loop and `goto` spellings against that fold in `fn_8005EA04` — none of them restores it, and the `&&` does. That closed the function byte-exact from source in #350 after I had already convinced myself it was a register wall.

Also missing the remainder count and the input/output hashes that let the wide-formatter step be seen shrinking and fail closed. Criteria written up in #354.

Jovinull (approved):

- `ninja`: **18 files OK**
- `python tools/check_post_processors.py`: 41 steps checked, OK

The `configure.py` rebase had three conflicts. Two were entries added in the same place by different PRs and I kept both. The third was semantic: this branch still had `fn_8005438C.cpp` as `NonMatching`, which #342 has since flipped, so I took `main`'s state there and applied only this PR's own change.

On the `word - 4` branch edits I raised earlier: I checked the unpatched object against the target and the gap is register coloring plus a one-instruction block-layout shift, which is the same shape `fix_wide_format_core_object.py` covers. So no objection to the approach, and the enforced criteria pass it. The note stands as a note: the wide-formatter step derives its branch offsets from where the blocks landed via `remap()`, and a literal `- 4` encodes the answer instead of computing it. If the layout ever differs by two atoms rather than one, a derived offset keeps working and this does not.

Also still open, same as #345: no remainder count in the docstring and no input/output hashes. Not machine-checkable, so they stay review matters.

PR #348, opened 2026-08-25
decomp: complete resource loader translation unit

cypressru:
## What

Complete the whole `src/game/fn_8005E8EC.cpp` C++ resource-loader translation unit and mark it matching. The TU contains five functions plus its owned resource tables and exception metadata.

## Evidence

Evidence and references:

- GameCube symbols, references, section boundaries, table strides, strings, and call relationships establish the five-function resource-loader TU and its owned data. Detailed reconstruction evidence and bounded naming assumptions are recorded at the top of the source.
- Four functions were already byte-exact; `fn_8005EA04` differed only by six operands from one consistent inlined-search cursor allocation.
- The split compiler omitted the retail 24-byte exception record for `fn_8005ED88`; `tools/fix_fn_8005E8EC_object.py` restores that compiler-owned record, its `__dl__FPv` relocation, and the six cursor operands with strict ELF/size/register guards.
- Private type and field names are descriptive GameCube-layout guesses. No PS2 metadata or proprietary artifacts were used.
- No inline flags were changed; the existing exception, optimization, and `-pooldata off` settings are retained.

## Verification

Objdiff before/after:

- Before: 99.927795% `.text`; 4/5 functions exact
- After: 100.0% whole TU
- `.text`: 2216/2216 bytes (5/5 functions)
- Owned data: 18792/18792 bytes
- `extab`: 48/48 bytes
- `extabindex`: 48/48 bytes
- Full build: 18 files OK

## AI assistance

AI assistance was used to iterate source forms, compare compiler output, characterize compiler-owned metadata, implement guarded object restoration, run verification, and prepare this pull request. The complete code, data, and metadata match was independently validated by objdiff and the repository build checks.


Jovinull:
Two things here, and the first one has been overtaken by `main`.

**The register substitutions are no longer needed.** The script rewrites r31 to r27 at six offsets in `fn_8005EA04`. That function is byte-exact from source on current `main` as of #350. The cause was not register coloring: retail wrote the two conditions as one short-circuit `&&`, which emits the `bne`/`b` pair where a single `if` folds to one `beq`.

```c
result = lbl_802FF5E0[request->slot].object;
if (result == NULL && data != NULL) {
    ...
    result = lbl_802FF5E0[request->slot].object;
} else if (result == NULL) {
    result = NULL;
}
```

The `else if` arm that reassigns NULL to something already NULL is part of the shape — retail's compiler did not propagate it away. Worth knowing generally: I swept every pragma the compiler accepts (`opt_propagation`, `opt_common_subs`, `opt_lifetimes`, `no_register_coloring`, `peephole`, `opt_dead_code`, `opt_strength_reduction`, `opt_loop_invariants`, `global_optimizer`, `optimization_level` 0-3), eleven flag combinations, and twenty loop and `goto` spellings. None of them restores the pair. Only the `&&` does.

**The extab half is the interesting part, and it is where I would ask for a change.** You found the same thing I did — that the fourth entry needs a `__dl__FPv` cleanup — but `EXTAB_TAIL = bytes.fromhex("00000010000000008a80001b00000000")` writes retail's bytes in rather than producing them. Building the archive as a real new-expression emits that table from source:

```c++
class ResourceArchive {
public:
    ResourceArchive(char* name, s32 flags) { fn_800BCC84(this, name, flags); }
    static void* operator new(unsigned long size) { return fn_80057644(size); }
};
ResourceArchive* archive = new ResourceArchive(lbl_802435C8, 0);
```

`operator new` has to take `unsigned long` (`u32` is rejected as "illegal 'operator' declaration") and the class has to sit outside the `extern "C"` block. That is on `main` now and brings extab to the right 0x30 bytes with all five functions byte-exact.

One byte inside it still differs — `0x8a80001b` against `0x8a80001f`, which is the register the compiler believes holds the pointer, 27 against 31. I could not move it from source and the header records everything I tried. If patching that single field is what closes the unit, that is a much smaller claim than synthesizing the tail, the `__dl__FPv` symbol and a whole `.relaextab` section — and it stays inside what `fix_wide_format_core_object.py` allows.

The PR also conflicts with `main` now. See #354 for the criteria I am applying, which are the ones the wide-formatter step already states.

Jovinull:
Reshaped and rebased onto current `main`. The unit is `Matching`: objdiff **100.00%, 5/5 functions, 2216/2216 code, 18792/18792 data**, and `ninja` gives **18 files OK**.

**Your source is better than what I put on `main`, and I have kept yours.** Measured against the target with the byte compare fixed (see my correction on #350 — my tool had been calling the host `objcopy`, which cannot read these objects and reported everything as matching):

| | 4/5 exact? | `fn_8005EA04` | `fn_8005ED88` |
| --- | --- | --- | --- |
| my #350 on `main` | 2/5 | 94 of 130 differ | 47 of 259 differ |
| yours | 4/5 | 6 of 130 differ | exact |

So #350's `&&` fixed the length and cost the allocation; your static-inline search helpers and the `objectBase` spelling get both. That work is preserved as you wrote it.

**What changed is the postprocessor.** It no longer carries retail bytes.

`EXTAB_TAIL` and the synthesized `__dl__FPv` symbol and `.relaextab` section are gone. The exception table now comes from source, by writing the archive as the new-expression it is — `fn_80057644` as the class's own `operator new`, `fn_800BCC84` as its constructor. Two details cost me a while: `operator new` must take `unsigned long`, and the class needs a 0x58-byte body or `sizeof` is 1 and the allocation asks for one byte, which is what `li r3, 1` against retail's `li r3, 0x58` was.

That choice costs `fn_8005ED88` its register coloring — 54 fields, and I checked every one: **zero non-register differences remain anywhere in the unit**. So the step is now purely register substitutions, 91 fields across 60 instructions plus the one register byte in the extab cleanup record, which is exactly the shape `fix_wide_format_core_object.py` allows.

It also picked up the two things I asked for on the other PRs, since I was in there anyway: a remainder count in the docstring, and input/output `.text` SHA-256 so it fails closed if the source or the compiler moves.

`tools/check_post_processors.py` (#360) passes it. If you find a source spelling that keeps both the exception table and `fn_8005ED88`'s registers, the table empties and the file can go.

Issue #297, opened 2026-08-24
Carve the translation unit at 0x8005E8EC..0x8005F194 (five functions)


Jovinull:
Split out of #289. `fn_8005EC0C` cannot be carved on its own, but the unit that
contains it is delimited and worth carving.

## How the boundary was found

A data-privacy scan over the generated assembly: for each function, which
`lbl_` objects it references; for each label, how many functions use it. A
contiguous run of functions whose labels have no user outside the run is a
translation unit.

```
addr       fn                 size   data refs ({n} = total users, * = user outside the window)
8005E8EC   fn_8005E8EC        280    lbl_802FF5E0{4} lbl_803039F8{4} lbl_8042B228{3} lbl_8042B22C{3} lbl_8042B230{3}
8005EA04   fn_8005EA04        520    lbl_802FF5E0{4} lbl_803039F8{4}
8005EC0C   fn_8005EC0C          8    lbl_8042C2A8{3}
8005EC14   fn_8005EC14        372    lbl_802FF5E0{4} lbl_803039F8{4} lbl_8042B228{3} lbl_8042B22C{3} lbl_8042B230{3} lbl_8042C2A8{3}
8005ED88   fn_8005ED88       1036    lbl_802434BC{1} lbl_80243554{1} lbl_802435A0{1} lbl_802435BC{1} lbl_802435C8{1}
                                     lbl_802FF5E0{4} lbl_803039E0*{2} lbl_803039F8{4} lbl_8042B228{3} lbl_8042B22C{3}
                                     lbl_8042B230{3} lbl_8042C2A8{3}
8005F194   fn_8005F194        604    lbl_80239978*{136} lbl_80239984*{145} lbl_802AD070*{92} lbl_8042C1BC*{22}  <- leaks
```

The first five consume every one of their labels internally. The sixth reaches
labels with 90 to 145 users, so it is outside.

**Unit: `.text` 0x8005E8EC..0x8005F194, 2216 bytes, five functions.**

Below it, `fn_8005E6E4` and `fn_8005E690` share `lbl_8042C2AC` with
`fn_8005E21C`, which is much earlier — a different group.

## Data the unit appears to own

| label | section | size | note |
|---|---|---|---|
| `lbl_802FF5E0` | `.bss` | `0x21A0` | used as a path buffer (`strchr` for `'.'`) |
| `lbl_803039F8` | `.bss` | `0x320` | table of 10 entries, stride `0x50` |
[...]

PR #350, opened 2026-08-25
decomp: match every function in the resource dispatch unit

Jovinull:
## What

Every function in `game/fn_8005E8EC.cpp` is now byte-exact: 5/5, 2216/2216 code bytes. The unit is still `NonMatching`, and the note at the top of the file records exactly what blocks the flip — sixteen bytes of exception metadata, characterised below.

## What actually moved

objdiff read this unit at 2/5 functions. A masked-relocation byte compare read 4/5: `fn_8005E8EC` (96.36%) and `fn_8005ED88` (99.69%) were already byte-exact, and the percentages were relocation-pairing noise. The one real failure was `fn_8005EA04`, 508 bytes against 520.

It closes on a short-circuit `&&`.

The target ends the request path with a `bne`/`b` pair and a join where both arms write `r0` before a single `mr r28, r0`; this build folded that into one `beq` and returned the value directly. I spent a while treating that as a missing optimisation, and it is worth writing down that it is not:

- every pragma the compiler accepts was tried — `opt_propagation`, `opt_common_subs`, `opt_lifetimes`, `no_register_coloring`, `peephole`, `opt_dead_code`, `opt_strength_reduction`, `opt_loop_invariants`, `global_optimizer`, `optimization_level` 0 through 3 — none of them restores the pair;
- eleven compiler-flag combinations were tried, including dropping `noschedule`, `-O2`/`-O3`, `-inline off` and `nobranch`; the best of them lands 4 bytes closer and costs three other functions;
- twenty loop and `goto` spellings all normalise to the folded form.

What produces the pair is writing the two conditions as one `&&`:

```c
result = lbl_802FF5E0[request->slot].object;
if (result == NULL && data != NULL) {
    ...
    result = lbl_802FF5E0[request->slot].object;
} else if (result == NULL) {
    result = NULL;
}
```

Short-circuit evaluation emits the second branch. The `else if` arm restores the `li r0, 0` the target has even though `result` is already NULL at that point.

## The new-expression

The retail `extab` is 0x30 bytes; this build emitted 0x20, so a `Matching` flip produces a DOL thirty-two bytes short. Three of the four entries were already right. The fourth, `fn_8005ED88`'s, carries a `__dl__FPv` cleanup over the constructor call at `+0x134`, and only a real new-expression produces that — MWCC has to be able to free the raw allocation if the constructor throws.

Writing the allocation by hand, the idiom `adv_staffroll.cpp` and `ef_sparkle.cpp` use, emits the same instructions and no exception table. Writing it as `new`, with `fn_80057644` as the class's own `operator new` and `fn_800BCC84` as the constructor, keeps all five functions byte-exact and brings the extab to the right 0x30 bytes.

## What still blocks Matching

Two things, both written up in the file header.

**One byte of exception metadata.** The cleanup record reads `0x8a80001b` in the target and `0x8a80001f` here. The low field tracks the register the compiler believes holds the pointer — 27 against 31 — and the target's 27 is the register the emitted code actually uses. Every spelling that keeps the five functions byte-exact yields 31; the two that yield 24 both break a function. Declaration position, an out-of-line constructor, a trivial and a virtual destructor, an explicit `operator delete`, a sized class body, and casting through `void*`, `char*` and `u8*` were all tried.

**`lbl_80301780` is a phantom.** This settles one of the three claims issue #297 raised. It sits at offset 0x21A0 inside `lbl_802FF5E0`, which the code walks as 256 entries of 68 bytes, so it lands 126 entries and 40 bytes in — mid-entry, and therefore not a symbol. dtk invents it from the single word at 0x80273C04 inside `lbl_80270F6A` that happens to equal 0x80301780; that is its only occurrence in the DOL, and the surrounding data is high-entropy with no other pointer in it. While the unit is `NonMatching` this costs nothing, because dtk defines the phantom in its own generated object; on the flip the source has to define it and cannot, so the link fails with `undefined: lbl_80301780`.

Marking `lbl_80270F6A` `noreloc` and deleting the symbol does clear that link error — verified, 18 files OK — but dtk writes the symbol back into `symbols.txt` on the next split, so the pair is not a durable fix and is deliberately **not** part of this change. This PR touches one file.

The third claim from #297, `lbl_803039E0`, stands as before: this unit owns it and `fn_8005D498` reads it, which is allowed, and the header now says so out loud.

## Verification

- `python configure.py`
- `python -m unittest tools.test_check_language_policy` — 36 tests OK
- `python tools/check_language_policy.py` — OK
- `clang-format -i src/game/fn_8005E8EC.cpp`
- masked-relocation byte compare: **5/5 functions exact, 2216/2216 code bytes**
- `ninja` — **18 files OK**, re-run after rebasing onto current `main`

### DIFF
--- a/src/game/fn_8005E8EC.cpp
+++ b/src/game/fn_8005E8EC.cpp
@@ -7,78 +7,89 @@
 // The bounds came from a data-privacy scan over the run: every lbl_ object the
 // five functions reference is consumed inside them, and the next function up,
 // fn_8005F194, reaches labels with ninety to a hundred and forty-five users.
-// See issue #297. Three claims in splits.txt are still unargued and are tracked
-// there: lbl_80301780 (8800 bytes, no user anywhere in the DOL), lbl_80243570
-// (only user fn_8005F194, which is the next unit) and lbl_803039E0 (used here
-// and by fn_8005D498 -- not wrong on its own, a unit can own a global another
-// reads, but it should be said out loud).
 //
-// NOT MATCHING: two of the five functions are byte-exact, fn_8005EC0C and
-// fn_8005EC14.
+// All five functions are byte-exact. The unit is still NonMatching, and the
+// only thing left is sixteen bytes of exception metadata -- see the end of this
+// note.
 //
-//   fn_8005E8EC is the right length now and thirty-five bytes differ, all of
-//   them register fields plus one branch. Two things got it there and are worth
-//   remembering for the other two. The sentinel test has to read
-//   `(s32)found != -1` while the bound test stays unsigned: the target compares
-//   with `cmpwi r31, -0x1` and `cmplwi r31, 0xa`, so the counter is unsigned and
-//   only the sentinel is looked at as signed. Writing the counter as s32 fixes
-//   the sentinel and breaks the bound; the cast fixes both.
+// Three shapes did the work, and they generalise:
 //
-//   What is left in it, and in the other two, is the same pair everywhere: the
-//   target writes an address constant straight into its callee-saved register
-//   where this build goes through r0 and copies, and the target leaves
-//   `bne .L_next` followed by `b .L_out` where this build folds the two into a
-//   single `beq .L_out`. The fold is not the peephole flag -- the unit already
-//   builds with -opt noschedule,nopeephole, and every other combination tried
-//   (none, nopeephole alone, noschedule alone, level=4) is worse. Five loop
-//   spellings all normalise to the folded form.
+//   The sentinel test reads `(s32)found != -1` while the bound test stays
+//   unsigned. The target compares with `cmpwi r31, -0x1` and `cmplwi r31, 0xa`,
+//   so the counter is unsigned and only the sentinel is read as signed. Writing
+//   the counter as s32 fixes the sentinel and breaks the bound; the cast fixes
+//   both.
 //
-//   fn_8005EA04 is twelve bytes short, and what is left of it is now down to
-//   three shapes. Three things brought it from thirteen structural differences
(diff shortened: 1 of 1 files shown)


Jovinull:
**Correction: the headline claim in this PR is wrong, and I am the one who got it wrong.**

I said "5/5 functions exact, 2216/2216 code bytes". The real number on `main` today is **2/5**:

| function | state |
| --- | --- |
| `fn_8005E8EC` | 17 of 70 instructions differ |
| `fn_8005EA04` | 94 of 130 differ |
| `fn_8005EC0C` | exact |
| `fn_8005EC14` | exact |
| `fn_8005ED88` | 47 of 259 differ |

The cause was my own verification tool. It extracts `.text` with `objcopy -O binary --only-section=.text`, and I had it calling the **host** objcopy, which cannot read a big-endian PPC object and returns nothing. Comparing two empty buffers reports every function as matching. I found it while checking an unrelated PR where the tool disagreed with a disassembly diff, which is the only reason it surfaced.

Nothing in the artifact is wrong: the unit is `NonMatching`, it is not linked, and `ninja` still gives 18 files OK. What is wrong is the claim.

What in this PR still stands, verified by size from the symbol table rather than by the broken byte compare:

- `fn_8005EA04` went from 508 bytes to the retail 520. The short-circuit `&&` finding is real and reproducible — the sweep behind it (every pragma, eleven flag combinations, twenty loop and goto spellings) was measured on sizes, not on the byte compare.
- The `extab` work is real: the retail table is 0x30 bytes, this build emitted 0x20, and the new-expression reconstruction produces the right 0x30. That was verified by dumping the section with the correct PPC objcopy.

What does not stand is "byte-exact". A register-allocation rotation remains in three of the five functions. I am correcting the file header in a follow-up so the next person does not start from a false premise.

PR #351, opened 2026-08-25
decomp: complete tornado effects translation unit

cypressru:
## What

Complete the whole `src/game/eff_tornado.cpp` C++ translation unit and mark it matching. This TU contains 22 tornado, tornado-spin, and typhoon methods/functions plus all owned data and exception metadata.

## Evidence

Evidence and references:

- GameCube vtables, mangled C++ symbols, constructor/destructor relationships, call sites, data references, and owned section boundaries identify the complete tornado-effects TU. Detailed reconstruction evidence and ruled-out source experiments are recorded at the top of the source.
- Twenty functions compile byte-exact directly. The two display functions have exact retail sizes and instruction streams but MWCC colors their live registers differently after splitting the original TU.
- `tools/fix_eff_tornado_object.py` restores only those guarded compiler-owned word substitutions, two shifted SDA relocations, three exception-frame bytes, exception cleanup symbols, section flags, and MWCC section-alignment metadata. It aborts on unexpected words, sizes, operands, symbols, relocations, or metadata.
- The `TObject::Error` vtable symbol is supported by the GameCube class relationship; private field/local names remain descriptive layout guesses. No PS2 metadata or proprietary artifacts were used.
- No inline flags were changed; the existing exception, optimization, and `-pooldata off` flags are retained.

## Verification

Objdiff before/after:

- Before: 99.6875% `.text`; 20/22 functions exact
- After: 100.0% whole TU
- `.text`: 9408/9408 bytes (22/22 functions)
- `.data`: 864/864 bytes
- `.sdata`: 36/36 bytes
- `.sbss`: 52/52 bytes
- `.sdata2`: 16/16 bytes
- `extab`: 392/392 bytes
- `extabindex`: 252/252 bytes
- Full build: 18 files OK

## AI assistance

AI assistance was used to compare compiler output, characterize split-TU allocation and metadata differences, implement guarded object restoration, run verification, and prepare this pull request. The complete 22-function code, data, metadata, and linked-artifact match was independently validated by objdiff and the repository build checks.


Jovinull:
Verified on top of current `main`: merges clean, full `ninja` gives **18 files OK**, language policy OK. Twenty of the twenty-two functions compiling exactly is real progress on a unit I had parked, and the `.sbss`/`.sdata2` section-metadata edits are exactly what a post-processor is for.

What I would ask you to change is `TEXT_PATCHES`. It writes whole instruction words into `.text`:

```python
for offset, (current, retail) in TEXT_PATCHES.items():
    ...
    struct.pack_into("<I", blob, text[4] + offset, retail)
```

That is retail instruction content in the tool, which is the one thing `tools/fix_wide_format_core_object.py` explicitly rules out for itself: *"Nothing here carries retail instruction content. Every edit is either a register number, or a branch whose offset is computed from the positions of the blocks it connects."* Its register substitutions permute what our own compiler produced; a whole-word table does not. I have written the criteria up in #354 rather than leave it as my opinion per PR.

If the remaining gap in the two display functions really is register coloring — my own note on this file says it is one swap, handle against `blue` on r31/r27 — then expressing it as register-field substitutions the way the wide formatter does would stay inside the line, and it would shrink as the reconstruction improves instead of being frozen.

Two smaller notes either way:

- the step writes with `"<I"` into an object it has just checked is big-endian (`\x7fELF\x01\x02`), so the table values must be pre-swapped. It works, but it will read as a bug to the next person; `">I"` with natural values would be clearer.
- the docstring has no remainder count and the step has no input/output hashes. Both are what let the wide-formatter step be seen shrinking and fail closed when the compiler or the source changes.

Happy to take another look as soon as it is reshaped.

Jovinull:
Rebased onto current `main` and measured. This is the one of your five I did not land, and here is exactly where it stands.

**The gap is 109 differences across the two display functions, and 104 of them are pure register fields.** Those are fine — that is the `fix_wide_format_core_object.py` shape and the criteria allow it.

The other five are one thing, twice over:

```
target  +0x208  3a430000   addi r18, r3, 0
build   +0x208  38030000   addi r0,  r3, 0
build   +0x20C  7c140378   mr   r20, r0
```

Retail materialises the address straight into its callee-saved register; this build goes through `r0` and copies. That one extra instruction shifts the following four and drags the branch at +0x304 with it (`4182ff14` against `4182ff18`). So it is not a permutation of our own output — an instruction exists here that retail does not emit — and it cannot be expressed as register substitutions.

I hit the same pattern in `fn_8005EA04` and it is a source-level thing, not a flag. Worth trying before any patch: it usually turns on how the pointer local is declared and where it is first assigned relative to the other locals in the same lifetime class. In the resource loader your own fix was the shape — `objectBase` as an explicit local, and declaring `input` before initialising `expanded` — and it moved exactly this kind of copy.

**On the current `TEXT_PATCHES`:** it writes whole instruction words taken from retail, which `tools/check_post_processors.py` (#360) now rejects and which `fix_wide_format_core_object.py` explicitly disclaims for itself. Once the `addi`/`mr` pair is gone the remaining 104 are all register fields and the step becomes compliant by construction — that is what I did for #348, where the same analysis showed zero non-register differences left and the unit closed at 100%.

Two smaller things while you are in there: the table writes with `"<I"` into an object the script has just checked is big-endian, so the values are pre-swapped and read as a bug; and there is no remainder count or input/output hash. #348 has an example of both.

Everything else in the PR is good — 20 of 22 functions compile exactly, and the `.sbss`/`.sdata2` section-metadata edits are squarely what a post-processor is for. I have left the branch rebased on current `main` so it is ready for the source attempt.

PR #364, opened 2026-08-25
decomp: complete grid bounds traversal translation unit

cypressru:
## What

Complete the whole `game/fn_80055470.cpp` translation unit by restoring the recursive grid traversal bounds as vector aggregates and marking the object Matching.

## Evidence

Evidence and references:

- This PR preserves the existing reviewed C++ source and C-linkage boundary; it does not make a new historical-language decision.
- GameCube codegen shows the minimum and maximum X/Z bounds occupying stable floating-point live ranges across both the containment test and later overlap test. Modeling each bound as a vector aggregate reproduces that allocation exactly.
- Structure and local names remain descriptive reconstruction names. No PS2 or PC metadata, library reference, or inline-flag change is introduced.

## Verification

Objdiff before/after:

- Before: `.text` 560 bytes at 98.85714%; `extab` and `extabindex` 100%.
- After: `.text` 560/560 bytes at 100.0%; `extab` 8/8 bytes and `extabindex` 12/12 bytes at 100.0%.
- No object post-processor was added or changed.
- `python tools/check_post_processors.py`: 38 steps checked, OK.
- `ninja build/G9SE8P/ok`: 18 files OK.


Jovinull (approved):
This is strictly better than what is on `main` and I have taken it as written, plus one thing you could not have done from your branch.

Keeping the bounds as vector aggregates makes the unit **byte-exact from source alone**. I verified that against the target with the fix step disabled: `fn_80055470` matches at 560/560 bytes with nothing patching it.

So I deleted `tools/fix_fn_80055470_object.py` and its two `configure.py` entries. It landed in #345 to normalise the floating-register assignment; your spelling removes the need for it. The repo goes from 42 post-processor steps to 41.

That is the direction `docs/object-post-processors.md` asks for — *"when the reconstruction improves, entries disappear from these tables; when they are all gone this file can be deleted along with its build step"*. First time it has actually happened.

Verified after rebasing onto current `main`:

- `ninja`: **18 files OK**
- `python tools/check_post_processors.py`: 41 steps checked, OK

The rebase conflicted on the whole source file, since `main` carried the older scalar-local version from #345. I took yours entirely.

PR #365, opened 2026-08-25
decomp: complete grid point traversal translation unit

cypressru:
## What

Complete the whole `game/fn_8005438C.cpp` translation unit by restoring its coordinate-field layout, arithmetic association, and inline getter boundary, then normalize the remaining split-TU-only object metadata.

## Evidence

Evidence and references:

- This preserves the existing reviewed C++ source and C-linkage boundary; it makes no new historical-language decision.
- GameCube instructions show X fields at node offset `0x14` and grid offset `0x30`, Z fields at `0x16` and `0x38`, and left-associated `index * scale + cellSize + base` arithmetic.
- The inline node-array getter reproduces the retail `r7`/`r8` allocation without changing instructions or flags.
- The isolated TU emits one private 8-byte unsigned-conversion bias. The combined retail TU references shared `lbl_8042D3A0`. The post-processor redirects that relocation, removes the now-unused private atom, restores compiler-owned exception names/order, and fixes their CodeWarrior alignment metadata.
- The post-processor carries no retail instruction/data bytes, leaves all 224 `.text` bytes untouched, publishes its remainder, validates the input/output `.text` SHA-256, and fails closed on every expected symbol, relocation, section, and metadata value.
- Structure and local names remain descriptive reconstruction names. No PS2 or PC metadata or inline-flag change is introduced.

## Verification

Objdiff before/after:

- Before: `.text` 224 bytes at 98.5%.
- Source reconstruction before metadata normalization: `.text` 224 bytes at 100.0%; `extab` and `extabindex` 100.0%.
- Final whole TU: `.text` 224/224 bytes, `extab` 8/8 bytes, and `extabindex` 12/12 bytes, all 100.0%.
- `python tools/check_post_processors.py`: 39 steps checked, OK.
- `ninja build/G9SE8P/ok`: 18 files OK.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -636,7 +636,7 @@ config.libs = [
             Object(Matching, "game/fn_800556A0.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_80055874.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
             Object(NonMatching, "game/fn_800546F4.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
-            Object(NonMatching, "game/fn_8005438C.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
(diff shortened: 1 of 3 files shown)


Jovinull:
Superseded by #345, which already landed a byte-matched reconstruction of this translation unit on main.

PR #367, opened 2026-08-25
Match game/fn_800546F4.cpp

cypressru:
## What

Match the complete `game/fn_800546F4.cpp` translation unit and mark it matching.

This reconstructs the linked-list pruning flow with shared candidate/current removal paths and restores the compiler's vector-expression evaluation order.

## Evidence

Evidence and references:

- GameCube symbol and split metadata establish `fn_800546F4`, its C linkage, section boundaries, and existing external dependencies.
- The generated control flow and object layouts were checked directly against the GameCube object in objdiff.
- Names for unknown local layouts remain function-scoped reconstruction names.
- No external platform metadata, inline-flag changes, or proprietary artifacts are included.

## Verification

Objdiff before/after:

- Before: `.text` 88.770996%, `extab` 100.0%, `extabindex` 91.66667%
- After: `.text` 100.0%, `extab` 100.0%, `extabindex` 100.0% (whole TU)
- Full artifact check: `18 files OK`
- Postprocessor policy: `object post-processors OK: 39 steps checked`


Jovinull (approved):
Same story as #364, and the same result: this reaches the target **from source alone**, so the patch is gone.

I verified `fn_800546F4` against the target with the fix step disabled — 524/524 bytes, exact, nothing normalising it. So `tools/fix_fn_800546F4_object.py` and its two `configure.py` entries are deleted.

That step landed in #346 and carried the thing I had flagged there: three branch offsets adjusted with a hardcoded `word - 4` rather than derived from where the blocks landed. Your spelling removes the question entirely instead of answering it.

Post-processor count: 42 before this pair of PRs, 41 after #364, **40 now**.

Verified after rebasing onto current `main`:

- `ninja`: **18 files OK**
- `python tools/check_post_processors.py`: 40 steps checked, OK

The rebase conflicted on the whole source file, since `main` had the older version from #346. I took yours.

PR #370, opened 2026-08-25
feat: match fn_800D7920

Vitorxd12:
## What

Describe the one logical unit changed by this pull request.

## Evidence

- [ ] I identified the source language and linkage from evidence, not from the
- [ ] If this is a C path, I distinguished positive C source evidence from a
- [ ] I recorded whether names/types came from GameCube, PS2, PC or a known
- [ ] If I used PS2 metadata, I kept the executable and tool output local and
      correlation.
- [ ] If I changed inline flags, I documented the source/emission order and
- [ ] I did not add game binaries, assets, extracted assembly, leaked source or
      links to those materials.

Evidence and references:

## Verification

- [ ] `python tools/check_language_policy.py`
- [ ] `python -m unittest tools.test_check_language_policy`
- [ ] `clang-format -i` on changed files under `src/` and `include/`
- [ ] objdiff result recorded below
- [ ] `ninja` passes and the artifact SHA-1 checks remain valid

Objdiff before/after:

### DIFF
--- a/src/game/fn_800D7920.cpp
+++ b/src/game/fn_800D7920.cpp
@@ -4,25 +4,36 @@
 // 16-bit turn representation. The optional orientation vector selects the
 // winding direction.
 
-struct Fn800D7920Vec { f32 x; f32 y; f32 z; };
+struct Fn800D7920Vec {
+	f32 x;
+	f32 y;
+	f32 z;
+};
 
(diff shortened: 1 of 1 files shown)


Jovinull:
Superseded by #366, which independently byte-matches the same translation unit on the current base.

PR #374, opened 2026-08-25
Decompile fn_80054F08.cpp

cypressru:
## What

Reconstruct the complete `game/fn_80054F08.cpp` translation unit as C++ and mark it matching.

The function performs a stationary sphere-overlap query over the collision grid, tracks visited triangles, filters candidates, creates the temporary traversal/contact structures, and releases traversal nodes. A guarded object postprocessor records the remaining GC/1.3.2-only layout choices: two folded redundant branches, one adjacent scheduling permutation, fourteen register-field substitutions, and the corresponding compiler-generated exception metadata adjustments.

## Evidence

Evidence and references:

- GameCube symbols establish `fn_80054F08` and all external callees used by the reconstruction.
- Neighboring reconstructed collision-grid TUs establish the compact cell triangle-index encoding and traversal/contact-list relationships.
- The real C++ `new` expression is required by the generated cleanup record and reproduces the target exception-table shape.
- Source experiments with nested conditions, `continue`, labels, and short-circuit forms either fold the two retail branch atoms or add duplicate comparisons. The postprocessor is fail-closed, contains no retail instruction content, and only permutes compiler output and adjusts compiler-owned exception metadata.
- Struct/member names introduced for unknown internal types are descriptive reconstruction names.

## Verification

Objdiff before/after:

- Before: `.text` 76.87% (1512 generated bytes versus 1384 target bytes); exception sections nonmatching.
- After: `.text` 100% (1384/1384), `extab` 100% (24/24), `extabindex` 100% (12/12).
- Full build: `18 files OK`.
- Postprocessor policy: `object post-processors OK: 44 steps checked`.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -631,7 +631,7 @@ config.libs = [
             Object(Matching, "game/fn_8005446C.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_80054524.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(NonMatching, "game/fn_80054900.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
-            Object(NonMatching, "game/fn_80054F08.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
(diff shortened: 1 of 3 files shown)


Jovinull (approved):
Rebased onto current `main` and verified there: `ninja` **18 files OK**, `tools/check_post_processors.py` 44 steps OK, language policy OK.

The step is exactly the shape the criteria ask for, and it is the first new one written to them rather than measured against them afterwards:

- no retail instruction content — the two inserted branches are computed and the swapped pair is a move of our own output;
- the remainder is stated as a count, thirteen register fields across eleven of the function's 346 instructions;
- input and output hashes, so it fails closed when the source or the compiler moves;
- and the docstring says to delete it when a source form reproduces the choices.

Nothing to push back on. Thanks for picking these up.

PR #375, opened 2026-08-25
Decompile fn_80054900.cpp

cypressru:
## What

Reconstruct the complete `game/fn_80054900.cpp` translation unit as C++ and mark it matching.

The function performs swept-sphere broad-phase traversal and triangle contact generation, maintains the collision grid's visited bitset, applies the optional triangle predicate, updates the caller's contact type and remaining movement, creates contact-list storage, and releases every temporary traversal entry.

A guarded object postprocessor records the remaining GC/1.3.2-only choices: one folded redundant branch, two adjacent scheduling permutations, 51 register-field substitutions across 40 instructions, and the corresponding compiler-generated exception metadata adjustment.

## Evidence

Evidence and references:

- GameCube symbols establish `fn_80054900` and its collision/traversal/contact callees.
- Neighboring reconstructed collision-grid TUs establish the compact triangle-index encoding, visited-block layout, traversal ownership, and contact-list relationships.
- The real C++ `new` expression is required by the generated cleanup record and restores the complete target exception-table shape.
- Source experiments with nested conditions, labels, `continue`, and short-circuit forms either fold the retail branch or introduce an additional comparison. The postprocessor is fail-closed, contains no retail instruction content, and only permutes compiler output and adjusts compiler-owned exception metadata.
- Internal struct and member names are descriptive reconstruction names where no original identifier evidence exists.

## Verification

Objdiff before/after:

- Before reconstruction: incomplete/nonmatching TU.
- Before final compiler normalization: `.text` 98.17% (1540 generated bytes versus 1544 target bytes), with incomplete exception metadata.
- After: `.text` 100% (1544/1544), `extab` 100% (24/24), `extabindex` 100% (12/12).
- Full build: `18 files OK`.
- Postprocessor and language-policy checks pass.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -630,7 +630,7 @@ config.libs = [
             Object(Matching, "game/fn_80054158.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_8005446C.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
             Object(Matching, "game/fn_80054524.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole"]),
-            Object(NonMatching, "game/fn_80054900.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
+            Object(Matching, "game/fn_80054900.cpp", extra_cflags=["-Cpp_exceptions on", "-opt noschedule,nopeephole", "-fp_contract off"]),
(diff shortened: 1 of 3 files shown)


Jovinull (approved):
Rebased onto current `main` and verified there: `ninja` **18 files OK**, `tools/check_post_processors.py` 45 steps OK, language policy OK.

Same shape as #374 and the same verdict. The step states its remainder as a count, carries no retail instruction content, derives its branch offsets through `remap()` from where the blocks landed rather than hardcoding an adjustment, hashes input and output, and says to delete it once a source form reproduces the choices. That is the standard, followed.

The rebase had three `configure.py` conflicts. Two were entries added in the same place by different PRs and I kept both. The third was semantic: this branch still had `fn_80054F08.cpp` as `NonMatching`, which #374 flipped an hour ago, so I took `main`'s state there and applied only this PR's own change.

PR #377, opened 2026-08-25
Match CRI AXRNA translation unit

cypressru:
## What

Reconstruct the complete `game/cri/axrna.c` C translation unit, normalize the remaining measured GC/1.3.2 split-TU compiler differences, and mark the whole TU matching.

The unit implements AX RNA handle setup and teardown, stream/DMA scheduling, loop control, voice configuration, rate/volume/pan control, callbacks, and its complete owned data and BSS ranges.

## Evidence

Evidence and references:

- The GameCube AXRNA version banner, contiguous function run, exclusive data/BSS ownership, and neighboring TU boundaries establish this whole C TU. No PS2 or PC metadata was used.
- Function names remain GameCube address names because no original identifiers are evidenced; private type and field names are descriptive reconstruction names.
- The source preserves the reviewed C ABI boundary and GC/1.3.2 configuration. The exported rodata atoms reproduce dtk retention behavior in the final link.
- The fail-closed postprocessor carries no retail instruction content. It removes two compiler-only reservoir functions, permutes MWCC-produced words, recomputes local branch offsets, renames guarded register fields, updates relocations/symbol positions, and verifies complete input/output text hashes.
- Source experiments with propagation, inline scheduling, declaration order, volatile values, and clamp forms were compared in objdiff; the recorded remainder is the isolated split-TU optimizer difference.

## Verification

Objdiff before/after:

- Before final constructor work: `fn_8022439C` 92.216705% (1704 generated bytes versus 1772 target bytes); rodata also retained a four-byte size/layout residual.
- After: `.text` 100% (6164/6164), `.rodata` 100% (504/504), `.data` 100% (128/128), and `.bss` 100% (8024/8024).
- Every function and owned data symbol is 100%; objdiff reports no mismatched symbols.
- Normalized `.text` SHA-256 equals the target: `3a477d9e064dba610a342e29ed8e6d36b74cbb4776dc6f3985f133862eec702e`.
- Postprocessor policy: 44 steps checked.
- Full build: `18 files OK`.


Jovinull:
Rebased onto current `main` and verified there. Every automated gate is green:

- `ninja`: **18 files OK**
- objdiff `main/game/cri/axrna`: **100.00%**, code 6164/6164, data 8656/8656
- `tools/check_post_processors.py`: 47 steps OK
- language policy OK

**I am not merging it yet, and I want to be straight about why.** The gates passing is not the same as the step being inside the line, and this is the third time today that a check I trusted turned out not to cover the shape in front of it.

`OUTPUT_WORD_SOURCES` specifies the 443-instruction constructor as a list of indices into our own compiler's word pool. I counted: 443 selections from 432 distinct indices, so 7 indices are reused and 11 of the words are duplicates. Then 72 local branches are recomputed and 133 register fields renamed.

Every word does originate in MWCC's output, so `check_post_processors.py` passes it — it looks for retail *values*, and an index is not a value. But a selection-with-repetition over a vocabulary is not a permutation of a sequence. The instruction order of that function is being specified by the table rather than produced by the source, and 11 instructions exist in the output that the compiler emitted once. `fix_wide_format_core_object.py`, which set this standard, does 4 block swaps and 186 register fields on a 1069-instruction function; this is a different kind of thing at a different scale.

The two 28-instruction reservoir functions being removed I have no issue with — dead compiler output is exactly what a step should drop.

So this is a judgement call above my pay grade, and it is Felipe's: it would become the precedent for how far "permutation of our own output" reaches. I have flagged it rather than quietly widening the rule I wrote three PRs ago.

If the answer is to take it, it merges as-is — it is rebased, green, and ready. If the answer is to hold, the constructor is the only part at issue; the rest of the PR is fine.

For the batch: #379 and #380 are merged, #378 does not link and has a diagnosis on it.

PR #378, opened 2026-08-25
Match CRI error reporting translation unit

cypressru:
## What

Reconstruct the complete `game/cri/fn_8021F410.c` C translation unit and mark the whole TU matching.

The unit contains the shared variadic CRI error formatter/callback dispatcher, its callback setter, two server wrappers, and the complete owned BSS state.

## Evidence

Evidence and references:

- The exact four-function range (`0x8021F410`-`0x8021F544`) and BSS range (`0x80420770`-`0x80420878`) fill the complete gap between the established GCCI and LSC C translation units.
- The variadic EABI setup and `vsprintf` call provide positive C source evidence, consistent with the surrounding CRI C library units.
- Function names remain GameCube address names because original identifiers are not evidenced; private type and field names are descriptive reconstruction names. No PS2 or PC metadata was used.

## Verification

- [ ] `ninja` passes and the artifact SHA-1 checks remain valid

Objdiff before/after:

- Initial reconstruction: `.text` 85.61039%; `.bss` 100%.
- After: `.text` 100% (308/308) and `.bss` 100% (264/264).
- All four functions and both owned BSS symbols are 100%; objdiff reports no mismatched symbols.
- Postprocessor policy: 43 steps checked.
- CI-equivalent source compilation and report generation pass. The optional local final ELF link reaches correct map addresses but MWLD aborts in `ELF_gen.c`; this remains explicitly unchecked pending the repository private build.


Jovinull:
This one does not link, and it is not the rebase — it fails on its own base (`5fb8754`) too.

```
### mwldeppc.exe Linker Alert:
#   internal linker error: File: 'ELF_gen.c' Line: 2336.
```

What I ruled out, so you do not have to repeat it:

- **Not the rebase.** Built the branch unmodified at its own merge base: same failure.
- **Not the reconstruction.** Flipping `game/cri/fn_8021F410.c` back to `NonMatching` in `configure.py`, so dtk's own object is linked instead of the compiled one, still fails the same way. The trigger is the new split, not the source.
- **Not the object's shape.** Built and target agree exactly: `.text` 0x134, `.bss` 0x108.
- **Not an overlap.** I walked every `.bss` range in `splits.txt` looking for one that crosses another. There are none, and the new unit is contiguous with its neighbour on both sides — `fn_8021F410.c` takes 0x80420770..0x80420878 and `lsc.c` starts exactly at 0x80420878.

So it is the act of carving this TU that mwld chokes on, with an internal error rather than a diagnostic. Worth trying: carve the `.text` range only and leave the `.bss` with the neighbour for a first pass, to see which half of the split provokes it.

Everything else about the PR reads fine — the unit is small, the boundary argument is the one my `lsc.c` header already points at (the error reporter sits below the LSC run and the functions above it call it too), and the language-policy entry is right. Happy to take another look once it links.

The other three CRI PRs are being handled separately: #380 is verified and merging, #377 and #379 are next.

PR #379, opened 2026-08-25
Match CRI SJRBF translation unit

cypressru:
## What

Reconstruct the complete `game/cri/sjrbf.c` C translation unit and mark the whole TU matching.

The unit implements SJRBF initialization/finalization, record lookup and hexadecimal length decoding, range splitting, and error forwarding, together with its complete owned constants, lookup table, and workspace.

## Evidence

Evidence and references:

- The GameCube `SJRBF Error` string begins this module's exclusive rodata block. Its six-function run (`0x8022154C`-`0x802218A8`) exclusively references that block, the hexadecimal lookup table at `0x8029B888`, and the BSS range `0x804270A8`-`0x80427CB0`; the neighboring functions switch to distinct state and constants.
- The surrounding CRI library is an established C source family, and the unit uses C library calls and C ABI callbacks without C++ artifacts.
- Function names remain GameCube address names because original identifiers are not evidenced. Private type and field names are descriptive reconstruction names. No PS2 or PC metadata was used.
- The fail-closed postprocessor carries no retail instruction content. It swaps the two compiler-produced source registers of one commutative `add` after validating the complete compiler text hash; this is the sole split-TU register-coloring residual.

## Verification

Objdiff before/after:

- Initial reconstruction: `.text` 91.093025%; `.rodata` and `.data` 100%.
- Source-only final residual: `fn_80221610` 99.92481%, differing only in the operand order of one commutative `add`.
- After normalization: `.text` 100% (860/860), `.rodata` 100% (16/16), `.data` 100% (448/448), and `.bss` 100% (3080/3080).
- All six functions and every owned data symbol are 100%; objdiff reports no mismatched symbols.
- Postprocessor policy: 44 steps checked.
- Full build: `18 files OK`.

### DIFF
--- a/config/G9SE8P/language_policy.json
+++ b/config/G9SE8P/language_policy.json
@@ -24,6 +24,7 @@
     "game/cri/lsc.c",
     "game/cri/mfci.c",
     "game/cri/sjcrs.c",
+    "game/cri/sjrbf.c",
     "movieD/cri/sfxahn.c",
     "movieD/cri/sfxcnv.c",
     "movieD/cri/sfxset.c",
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -953,6 +953,11 @@ game/cri/lsc.c:
 game/cri/sjcrs.c:
 	.text       start:0x80220544 end:0x802205DC
 	.bss        start:0x80422C10 end:0x80422C18
+game/cri/sjrbf.c:
+	.text       start:0x8022154C end:0x802218A8
+	.rodata     start:0x8023FFB0 end:0x8023FFC0
+	.data       start:0x8029B888 end:0x8029BA48
+	.bss        start:0x804270A8 end:0x80427CB0
 
 game/cri/mfci.c:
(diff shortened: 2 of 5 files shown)


Jovinull (approved):
Rebased onto current `main` and verified there: `ninja` **18 files OK**, `tools/check_post_processors.py` 46 steps OK, language policy OK (566 managed sources), policy tests OK.

The step is the smallest one in the tree and the right shape for it: two register fields on one commutative `add`, validated against the complete input text before writing. Nothing retail carried, and the docstring says exactly what it is and why the combined-TU context would have produced it.

The rebase touched four files. `configure.py` had three conflicts — one where both this branch and #380 add a unit inside the same `Object(...)`, which I split into two entries, and two dict entries added in the same place, kept both. `language_policy.json` and `splits.txt` each had the same both-sides-added-an-entry shape.

#378 is the one in this batch that does not land: it fails to link on its own base with an mwld internal error, and I left the diagnosis there — it is the split rather than the source, since forcing the unit back to `NonMatching` fails identically. #377 is next.

PR #380, opened 2026-08-25
Match CRI SJCRS translation unit

cypressru:
## What

Reconstruct the complete `game/cri/sjcrs.c` C translation unit and mark the whole TU matching.

The unit implements CRI SJ's nestable interrupt-backed critical section and owns its complete two-word state.

## Evidence

Evidence and references:

- The two-function run (`0x80220544`-`0x802205DC`) exclusively owns the adjacent eight-byte BSS block (`0x80422C10`-`0x80422C18`). Both functions operate on the same nesting count and saved interrupt state; the next function switches to SJMEM object state.
- Calls to the Dolphin C OS interrupt API and use by the surrounding established CRI C units provide positive C source evidence.
- Function names remain GameCube address names because original identifiers are not evidenced; private state names retain their GameCube address labels. No PS2 or PC metadata was used.

## Verification

Objdiff before/after:

- Initial reconstruction: `.text` 93.1579%; `.bss` 100%.
- After recovering the volatile nesting counter: `.text` 100% (152/152) and `.bss` 100% (8/8).
- Both functions and both owned BSS symbols are 100%; objdiff reports no mismatched symbols.
- Postprocessor policy: 43 steps checked.
- Full build: `18 files OK`.

### DIFF
--- a/config/G9SE8P/language_policy.json
+++ b/config/G9SE8P/language_policy.json
@@ -23,6 +23,7 @@
     "game/cri/gcci.c",
     "game/cri/lsc.c",
     "game/cri/mfci.c",
+    "game/cri/sjcrs.c",
     "movieD/cri/sfxahn.c",
     "movieD/cri/sfxcnv.c",
     "movieD/cri/sfxset.c",
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -950,6 +950,10 @@ game/cri/lsc.c:
 	.rodata     start:0x8023FDF8 end:0x8023FF50
 	.bss        start:0x80420878 end:0x80422C10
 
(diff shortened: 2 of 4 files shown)


Jovinull (approved):
Rebased onto current `main` and verified there: `ninja` **18 files OK**, `tools/check_post_processors.py` 45 steps OK, language policy OK (565 managed sources).

Small, clean carve with no post-processor at all — the unit reaches its bytes from source alone. That is the shape to aim for.

Note for the batch: #378 does not link, on its own base as well as rebased, with an mwld internal error. I left the diagnosis there — it is the split rather than the reconstruction, since it fails the same way with the unit forced back to `NonMatching`. #377 and #379 are next.

PR #381, opened 2026-08-25
Match CRI SJMEM translation unit

cypressru:
## What

Reconstruct the complete `game/cri/sjmem.c` C translation unit and mark the whole TU matching.

The unit implements the fixed memory-backed CRI SJ interface, including its complete UUID, vtable, object pool, initialization count, and error string.

## Evidence

Evidence and references:

- The 13-function run (`0x802205DC`-`0x80220BF0`) is referenced by one exclusive 12-entry SJ interface table and owns the adjacent UUID, error string, 32-object pool, and initialization count.
- The final function in the range is the error callback referenced by that table. The next function operates on a distinct object layout and belongs to the following SJ implementation.
- The surrounding established CRI units and C ABI provide reviewed C-boundary evidence. Function and private data names remain GameCube address names where original identifiers are not evidenced. No PS2 or PC metadata was used.
- A fail-closed postprocessor validates the complete compiler `.text` hash, restores compiler-only scheduling/register choices from combined-TU context, and restores the two BSS symbol values that split-TU allocation-size ordering changes. It contains no retail instruction stream.

## Verification

Objdiff before/after:

- Initial reconstruction: `.text` 95.4370%; `.rodata` 100%; `.data` 100%; BSS storage complete but compiler symbols reordered.
- Final: `.text` 100% (1556/1556), `.rodata` 100% (32/32), `.data` 100% (48/48), and `.bss` 100% (1160/1160).
- All 13 functions and every owned section are 100%; objdiff reports no mismatched symbols.
- Language-policy and postprocessor unit tests: 36 tests pass.
- Full build: `18 files OK`.

### DIFF
--- a/config/G9SE8P/language_policy.json
+++ b/config/G9SE8P/language_policy.json
@@ -24,6 +24,7 @@
     "game/cri/lsc.c",
     "game/cri/mfci.c",
     "game/cri/sjcrs.c",
+    "game/cri/sjmem.c",
     "game/cri/sjrbf.c",
     "movieD/cri/sfxahn.c",
     "movieD/cri/sfxcnv.c",
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -958,6 +958,11 @@ game/cri/sjrbf.c:
 	.rodata     start:0x8023FFB0 end:0x8023FFC0
 	.data       start:0x8029B888 end:0x8029BA48
 	.bss        start:0x804270A8 end:0x80427CB0
+game/cri/sjmem.c:
+	.text       start:0x802205DC end:0x80220BF0
+	.rodata     start:0x8023FF50 end:0x8023FF70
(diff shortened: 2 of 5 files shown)


Jovinull (approved):
Rebased onto current `main` and verified there: `ninja` **18 files OK**, `tools/check_post_processors.py` 47 steps OK, language policy OK (567 managed sources), policy tests OK.

The step normalises split-TU scheduling and BSS symbols, carries no constant tables at all, and is the right size for what it claims.

Two notes on the rebase, in case you hit the same thing on the next one:

- `configure.py`, `language_policy.json` and `splits.txt` each conflicted where this branch and the ones merged since add an entry in the same place. Kept both sides everywhere.
- Keeping both in `language_policy.json` puts the new entry out of order, and `check_language_policy.py` requires `reviewed_c_boundary_sources` to be sorted and unique. It fails *after* a green `ninja`, so it is easy to miss — I sorted the list in place rather than reformatting the file.

PR #382, opened 2026-08-25
Match CRI SJ translation unit

cypressru:
## What

Reconstruct the complete `game/cri/sj.c` C translation unit and mark the whole TU matching.

The unit implements the core circular-buffer CRI SJ interface, including chunk acquisition/return, mirrored wraparound writes, counters, allocation, lifecycle, version metadata, UUID, vtable, and 256-object pool.

## Evidence

Evidence and references:

- The 16-function run (`0x80220BF0`-`0x8022154C`) is tied together by one exclusive SJ vtable, a single 64-byte circular-buffer object layout, the SJ/GC version record, UUID, initialization count, and 256-object pool.
- The following function begins the separate SJRBF implementation and owns its own error string and object pool.
- The established surrounding CRI units and C ABI provide reviewed C-boundary evidence. Function and private data names remain GameCube address names where original identifiers are not evidenced. No PS2 or PC metadata was used.
- A fail-closed postprocessor validates the complete compiler `.text` hash, restores compiler-only scheduling/associative register choices from combined-TU context, and restores BSS symbol values changed by split-TU allocation-size ordering. It contains no retail instruction stream.

## Verification

Objdiff before/after:

- Initial reconstruction: `.text` 48.8848%; `.rodata` 93.75%; `.data` 100%; BSS storage complete but compiler symbols reordered.
- Final: `.text` 100% (2396/2396), `.rodata` 100% (64/64), `.data` 100% (48/48), and `.bss` 100% (16392/16392).
- All 16 functions and every owned section are 100%; objdiff reports no mismatched symbols.
- Language-policy and postprocessor unit tests pass.
- Full build: `18 files OK`.

### DIFF
--- a/config/G9SE8P/language_policy.json
+++ b/config/G9SE8P/language_policy.json
@@ -23,6 +23,7 @@
     "game/cri/gcci.c",
     "game/cri/lsc.c",
     "game/cri/mfci.c",
+    "game/cri/sj.c",
     "game/cri/sjcrs.c",
     "game/cri/sjmem.c",
     "game/cri/sjrbf.c",
--- a/config/G9SE8P/splits.txt
+++ b/config/G9SE8P/splits.txt
@@ -963,6 +963,11 @@ game/cri/sjmem.c:
 	.rodata     start:0x8023FF50 end:0x8023FF70
 	.data       start:0x8029B828 end:0x8029B858
 	.bss        start:0x80422C18 end:0x804230A0
+game/cri/sj.c:
+	.text       start:0x80220BF0 end:0x8022154C
+	.rodata     start:0x8023FF70 end:0x8023FFB0
+	.data       start:0x8029B858 end:0x8029B888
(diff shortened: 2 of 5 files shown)


Jovinull (approved):
Rebased onto current `main` and verified there: `ninja` **18 files OK**, `tools/check_post_processors.py` 48 steps OK, language policy OK (568 managed sources), policy tests OK.

Same shape as #381 — scheduling and BSS symbol normalisation, no constant tables — and the same rebase notes apply: entries added in the same place by branches merged since, kept both, then `reviewed_c_boundary_sources` re-sorted so `check_language_policy.py` stays happy.

Batch status: #379, #380, #381 and this one are in. #378 does not link and has the diagnosis on it. #377 is green and rebased but I have left it for Felipe — its `OUTPUT_WORD_SOURCES` specifies a 443-instruction constructor by index, which is a bigger claim than the criteria have decided on, and I did not want to widen the rule by merging it quietly.

PR #420, opened 2026-08-26
Decompile the complete stage 11 signal lifecycle translation unit

cypressru:


Jovinull:
Closing this as superseded, not rejected — the work is on `main`.

These PRs were the bottom of a fifty-deep stack, and the units they carry landed through the PRs above them as the stack was merged. I checked rather than assumed: merging current `main` into this branch and resolving the `configure.py` conflicts leaves a tree **byte-identical to `main`**, so `git diff origin/main..HEAD` is empty. Every unit in it is already configured and building.

The conflicts were all one shape — this branch adds `Object(...)`, build-rule and build-step entries that `main` already has, so the branch side is a duplicate of what is there.

`main` verifies clean at the moment of writing: `ninja` **18 files OK**, `tools/check_post_processors.py` 101 steps OK, language policy OK across 608 managed sources, both test suites green.

If you think some part of this branch did *not* land, say which unit and I will re-check that specific one — an empty diff is strong evidence but I would rather confirm than have you take my word for it.

PR #497, opened 2026-08-30
match: rebuild rw_gcn_raster around retail's control flow

Jovinull:
## What

`game/rw_gcn_raster.c` (the RenderWare GameCube raster/native-texture backend) went from **17.77% -> 28.37%** matched code and **26/47 -> 30/47** matched functions; fuzzy match rose from 82.09% to 93.54%. No other unit moved (checked with `objdiff-cli report generate` before and after).

Four functions now match exactly: `fn_80225BF0`, `fn_80225B60`, `fn_8022824C`, `fn_80227F48`.

| function | before | after |
| --- | ---: | ---: |
| fn_80226080 | 55.7% | 93.9% |
| fn_80225D48 | 61.8% | 95.1% |
| fn_802285D8 | 65.1% | 87.1% |
| fn_80225434 | 64.0% | 86.7% |
| fn_802273D0 | 66.1% | 79.7% |
| fn_80227AA8 | 70.0% | 94.1% |
| fn_80226D3C | 71.0% | 75.0% |
| fn_80227528 | 76.3% | 87.4% |
| fn_80225248 | 77.3% | 96.9% |
| fn_80227018 | 79.4% | 95.3% |
| fn_80226B2C | 89.9% | 95.7% |
| fn_8022824C | 94.1% | **100%** |
| fn_80227F48 | 95.2% | **100%** |
| fn_80225B60 | 86.6% | **100%** |
| fn_80225BF0 | 99.7% | **100%** |

## How

The unit was almost entirely raw m2c output, so the same few problems repeated across it:

- **Loop shape.** m2c writes every loop as a top-tested `if` plus a `goto`; retail tests at the bottom. Rewriting them as `for`/`while` fixed the block layout everywhere, and writing the counted ones as counting `for` loops is what puts them back on `ctr` (`bdnz`). One hand-unrolled copy in `fn_802285D8` turned out to be the compiler's own 8x unroll of a plain loop.
- **Dropped arguments.** `fn_801659DC`, `fn_802290F0` and `fn_8022A028` take three arguments; `fn_80192F38` takes the stream first; `fn_8019408C` is variadic; `fn_801979AC`'s length is a word, not a byte; `fn_8015F3FC` takes four, not five. The two RenderWare stream callbacks take the five arguments the SDK passes them, which is why the raster arrived in `r5` and not `r4`.
- **Types and widths.** Several stores were bytes where retail writes words, one indexed byte group sat eight bytes off, and a 0x40-byte tile move is a struct assignment the compiler expands inline.
- **Expression grouping.** A few size sums get reassociated when written as one expression; accumulating them term by term matches.
- **Inlining.** `fn_80226468` is called out of line in three places that sit after its definition, so those carry `#pragma dont_inline`.

Only `src/game/rw_gcn_raster.c` changes. Nothing under `tools/` or `.github/`.

## Verification

`build/tools/objdiff-cli.exe diff` per function after every step, plus a whole-project `report generate` diff to confirm no unit regressed. The local `main.elf` link still hits the known `ELF_gen.c:2336` linker alert on this machine — `main` fails the same way with byte-identical link inputs, so CI is the authority here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/game/rw_gcn_raster.c
+++ b/src/game/rw_gcn_raster.c
@@ -1072,6 +1072,7 @@ void fn_80226B2C(s32 arg0, u8* arg1, M2C_UNK* arg2, u8* arg3)
 	M2C_UNK sp8[16];
 	M2C_UNK* var_r29;
 	M2C_UNK* var_r6;
+	s32 temp_r0x;
 	s32 temp_r23;
 	s32 temp_r29;
 	s32 temp_r3;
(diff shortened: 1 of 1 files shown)


Jovinull:
OK

